Issue #51, opened 2017-06-25
No easy way to add student users in bulk (M)


smurthys:
A script/procedure to add students from a CSV file (for example, a roster) will be helpful

smurthys:
For those with access to the [Gradebook repo](https://github.com/DASSL/Gradebook): The roster import process is probably a good template.

smurthys:
I have added a test roster file (PR #131) to help with testing the resolution to this issue.

The code part of the resolution should not work with just the sample roster file. Instead it should work with any reasonable roster. For example, it might be best to use `VARCHAR` (with no size) as the type of all fields.

s-aithal:
The sample roster file contains students from different programs belonging to different classes, and with different registration status. 

Is it safe to assume that the input file contains only those students that need to be added to ClassDB or should the import support filtering rows in the input file?

smurthys:
Import all rows without any filter.

PR #131, opened 2017-07-21
Add test Banner roster

smurthys:
The commits in this PR add a sample Banner CSV roster to facilitate testing of the eventual resolution to Issue #51 (add student users in bulk). A README file is also added.

The sample roster is added to the `tests` folder so it can become part of unit testing the code to add student users in bulk.

### DIFF
--- a/tests/importStudents/README.md
+++ b/tests/importStudents/README.md
@@ -1,4 +1,4 @@
-testBannerRosterREADME.md - ClassDB
+README.md - ClassDB
 
 Sean Murthy   
 Data Science & Systems Lab (DASSL), Western Connecticut State University (WCSU)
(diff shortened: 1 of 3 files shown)


afig:
Overall looks good. This will be definitely be helpful for eventually implementing a solution to above mentioned issue.

The only issue I noticed is that the filename on line 1 of the readme does not match the actual file name.

Another curiosity is that one of the rows contains an asterisk after each instance of "BS Computer Science". I am assuming it is intentional, and carries a meaning, but am not certain.

srrollo:
I also think it looks good, besides the one filename/comment Andrew pointed out.

smurthys:
Thanks for the review. I just pushed a commit renaming the file and correcting the error in the file prologue.

smurthys:
@afig: The asterisk in the file is generated by Banner. The asterisk has meaning outside the context of ClassDB, but I left it in there so the data is representative.

smurthys:
I will merge the changes tomorrow early morning EDT, just in case @s-aithal sees issues.

Issue #138, opened 2017-12-29
Can only track the last connection and DDL activity of students (M)


smurthys:
We need to be able to track all connection and DDL activities of all users: not just the last of these activities; not just for students.

Adding this feature requires:
- [ ] Detail table(s) to capture activity
- [ ] DDL trigger to insert activity to a detail table
- [ ] Log import to insert activity to a detail table
- [ ] Views/functions to return activities
- [ ] Function(s) to purge the detail table(s)

srrollo:
Some thoughts about this:
- A robust solution to this issue will probably intersect with many issues, or at least provide a logical opportunity to integrate their solutions.
- I image two tables could be used, one storing DDL activity, and one storing connections. I made a quick sketch of them [here](https://drive.google.com/file/d/1Ecl5f0eas6qbQXtsUBvQ1cQBmui0pVs2/view?usp=sharing) (please let me know if this link doesn't work).
- Any columns related to logging should then be removed from `ClassDB.student`.
- This would provide an opportunity to merge `ClassDB.student` and `ClassDB.instructor` (#49), since the two tables would be virtually identical at that point. The above link also contains a possible `User` table as a replacement.
- #137 would be indirectly solved by the removal of mentioned columns from `ClassDB.student`, as @smurthys mentions.
- `DDLCount` and `connectionCount` would no longer be calculated by the log import and DDL trigger. Instead, the relevant frequent view(s) would derive them from the Detail tables.

smurthys:
Thanks for the analysis @srrollo. Yes, in general, it may be good to stop working on issues whose resolutions may be voided by the new tables. Except, Issue #135 should be addressed so as not to interfere with new contributors' activities.

Things to think about in relation to the new tables:
- The `User` table needs a `fullName` attribute and `userName` should probably be renamed `loginName` or some such
- We need to think carefully about the PK for the detail tables: what do the time stamp fields represent? 
- Names of the time stamp fields may need revision
- What are the pros and cons of maintaining two separate detail tables? Would users have to execute two separate queries to get all user activities?

I think @KevinKelly25 and @baconbm will likely have some thoughts on the [sketch of the new tables](https://drive.google.com/file/d/1Ecl5f0eas6qbQXtsUBvQ1cQBmui0pVs2/view?usp=sharing) and their connection to views. Specifically, they might be able to draw the ER diagram and generate table schemas.

KevinKelly25:
[...]

PR #139, opened 2018-01-01
Add a rough draft of ERD

KevinKelly25:
Adds a ERD for the detail tables mentioned in issue #138


smurthys:
Thanks @KevinKelly25 for getting this started.

A few things to consider:
- The title of the HTML document needs revision. It presently says "Copy of ClassDB_Detail_Log_20171230", which is obviously not relevant
- Entity name `connectionActivity` needs to begin with upper case
- Some of the points enumerated in an [earlier comment](https://github.com/DASSL/ClassDB/issues/138#issuecomment-354604815) still need consideration  
- In general, I recommend looking through comments to Issue #138 to make sure all requirements are met (perhaps they are, but do check)

I wonder what people think about the naming convention for attribute names in tables: Thus far, ClassDB has made first-letter lower case, whereas Gradebook makes first-letter upper case. Which approach should we use? 

I prefer first-letter upper case because that seems to be the industry norm, and it distinguishes attributes from variables in code.

KevinKelly25:
Added corrections as discussed in today's meeting and issue #138 .

KevinKelly25:
Added corrections as discussed in today's meeting and issue #138 .

afig:
Thanks for making the changes @KevinKelly25, they look great. 

I have a few minor suggestions:

- The `StatementSucceededAtUTC` attribute in the `DDLActivity` entity should start with a lowercase letter to match the convention currently used elsewhere in the diagram and project. @smurthys made a comment earlier about starting a discussion related to this convention, but we have not yet committed to a decision.

- I believe that it would make more sense for the new file `ERSchema.html` to be renamed to `ER-Schema.html`. This follows the naming convention of other files in the docs folder, and matches the title located within the `<title>` tags in the HTML file.

- @smurthys [proposed](https://github.com/DASSL/ClassDB/issues/138#issuecomment-354604815) adding an attribution box similar to the one used in [Gradebook's ER diagram](https://github.com/DASSL/Gradebook/blob/2d6693589c96e4e904c41d074c486afe935af612/docs/DASSL_Conceptual_Schema.html). I agree that adding this information to the diagram would be beneficial.

---

In terms of naming conventions for attribute names in tables, I do not have a strong personal preference between the two major conventions. However, since first-letter-uppercase appears to be more common overall, then I agree that we should consider switching. If we agree to switch, then a separate issue should be opened.

srrollo:
Overall, the latest ER schema looks good. 

One additional observation I have is that `StatementSuceededAtUTC` is not the best name for this attribute. Currently, we are using `statement_timestamp` to determine the timestamp for DDL activity. According to [the Postgres docs](https://www.postgresql.org/docs/9.6/static/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT), this returns the timestamp when the Postgres server received the current statement from the client. Perhaps `StatementStartedAtUTC` would be a better name.

I also agree that beginning attribute names with capital letters is a better convention, and that this should be addressed in a future issue.

KevinKelly25:
Thank you @afig and @srrollo for the comments. I fixed the issues you brought up. I did leave the attributes in the current system we have, lower case first. However, I also agree that switching to what is the industry norm according to @smurthys is probably a good idea.

smurthys:
Good changes @KevinKelly25 

A few observations:
- Relationships need names
- The relationship cardinality on the side of `ConnectionActivity` should be `0..*`: a user is not required to have a connection. The cardinality on the other side of this relationship should be `1..1`.
- Partial ID attributes in weak entities should have a dashed underline, if that is possible in draw.io
- The HTML title should be `ClassDB ER Schema`, unless draw.io requires the title element's content to match the file name (I doubt that is the case)
- I recommend adding the text `ClassDB ER Schema` at the top of the diagram to make it appear like a "heading" (with sufficient space before and after the heading text)
  - The heading will be useful when the image version of the schema is embedded in the docs
- I recommend moving the attribution box to the bottom of the diagram with its left aligned with the left of entity `User`
- Names in the attribution box (and in general in any artifact) should reflect people who have made "meaningful contribution". IIRC, that should be Kelly, Rollo, and Murthy, but please feel free to correct me
   - Past contributors are familiar with attribution guidelines; new contributors are yet to be appraised

smurthys:
I have added the Issue #140 about letter case. Please discuss that specific issue over there.

Because the ER Schema needs to be edited, it is OK if the issue is fixed in this schema. However, I am also OK if it is addressed in the next round (because we know this schema will soon be outdated). I'd prefer if this PR is approved and merged before this afternoon's call (but of course only without sacrificing quality).

KevinKelly25:
Updated ERD with corrections for @smurthys comments and used new standard for attribute names discussed in issue #140

KevinKelly25:
Updated the Attribute box to include authors with meaningful contributions as per the discussion in the meeting today.

smurthys:
Good work @KevinKelly25.

After all reviewers have approved, please hit the "Merge pull request" button to merge your changes to master. I believe the changes should merge without any conflict.

Remember to delete your branch **after** the merge is complete **and** you have verified the file appears in the master branch.

afig:
Also of note, when merging a pull request, there are three options, which can be seen by clicking the drop-down arrow next to the merge button. The three options are:

- Create a merge commit
- Squash and merge
- Rebase and merge

So far, we have been using the "Create a merge commit" option, and thus this Pull Request should also do the same. "Create a merge commit" should be the default options, but is worth checking. If you wish, you can read more about the differences between each method [here](https://help.github.com/articles/about-merge-methods-on-github/).

KevinKelly25:
Thank you, once approved by @baconbm I will do the create a merge commit

PR #145, opened 2018-01-06
Add-M2-ERD

KevinKelly25:
Added new entities and attributes to match the desired M2 goal as described in the [M2 wiki](https://github.com/DASSL/ClassDB/wiki/User-and-Team-Management-(M2))


KevinKelly25:
Fixed two syntax errors in ERD.

smurthys:
Good start @KevinKelly25 

In addition to the issue @srrollo mentions:
- Attribute `UserType` is extra. It should be removed
- Attribute `SchemaName` is missing in `User`
- Attribute `IsInstructor` is misspelled: missing `r`
- It is better to show the identifying relationships using horizontal lines (instead of angled lines). Doing so requires more space between entities, and that is OK
- In the attribution box, add a line break after DASSL and show DASSL's URL on the next line (as is done in code files)
- We need to add a "Notes" box which will also be dash-lined box like the attribution box. The notes should contain the following (each item on its own line) :
   - The ER diagram is drawn using https://draw.io and employs a simple UML-like syntax
   - Solid underline denotes ID attributes; dashed underline denotes partial-ID attributes
   - Asterisks mark required attributes; brackets denote derived attributes
   - A Role may be either a user or a team, but not both.

smurthys:
I also have the following suggestions:
- Rename the ID attributes of `User` and `Team` to `UserName` and `TeamName` respectively.
- Rename the relationships `Is a` to `Represents` to avoid triggering the thought of inheritance. (The "Extended ER Model" does support inheritance, but I think we should keep the schema simple for now.)

KevinKelly25:
Made corrections for ERD. I made a "Notes" box like the attribution box. However, I am not sure the placement is the best area for it. The reasoning for the current placement is so that the notes box is inherently read first since it explains the format of the ERD.

smurthys:
Nice changes @KevinKelly25. Like @baconbm, I agree the notes box is situated perfectly.

A few things to consider (I know, obsession about details 💯):
- Make `ConnectionCount` the last but one attribute in `User`
- Rename relationship `Creates` to `Causes`
- The lines of relationship `Creates` appear wider apart than the lines of relationship `Performs`
- Rephrase the last note as "A role may represent either a user or a team, but not both"
- I wonder if the notes box should include a note about double-lined rectangles denoting weak entities and double-line relationships being identifying relationships. If added, it should be as the fourth note.

The last note in the box raises an interesting question about the cardinality constraint across from `Role` in each `Represents` relationship: Should that constraint should be `0..1` instead of `1..1`, as in "a role represents 0 or 1 team" and "a role represents 0 or 1 user"?

What do you all think?

KevinKelly25:
@smurthys That does make sense to make it so a role is 0..1 for both since you can be one and not the other. However, it does bring up two questions.

1) Can a role be both a team and a user? I imagine that a role should be one or the other but not both.

2) Should user and team be connected as well? Since user(s) make up a team.

smurthys:
@KevinKelly25 Perhaps I misunderstand your question, but I believe the last note says the role can only represent a user or a team, but not both.

We should relate team and user when we support team functionality. No need to depict the details of something as yet unsupported.

KevinKelly25:
@smurthys I worded it wrong and meant it towards how we have ClassDB currently but I looked more closely and found out the answer. Thank you for the response.

afig:
I agree that a cardinality constraint of `0..1` from `Role` makes more sense. Adding the additional notes about the syntax we're using sounds good as well.

One <sup>tiny</sup> fix: The line between the Role and Team boxes is not fully connected on both ends to the edge of the boxes. Same with the line from Role to User on the Role side.

- With draw.io, you can zoom in and out by pressing the `ctrl` key and scrolling with the `mousewheel` while your cursor is within the diagram.
- If grid-snapping is causing you grief, you can temporarily disable it in settings, or by holding down the `alt` key after "grabbing" the point to move.

KevinKelly25:
@afig Thank you for the tips on Draw.IO. They definitely will help as the grid snapping was making things harder.

KevinKelly25:
Made corrections for issues brought up by @smurthys  and @afig

smurthys:
The last note should read "Entities in double-lined rectangles are weak entities; double-lined relationships are identifying relationships"

Thinking further about how best to avoid confusion for the reader, and being precise, I feel the `Role` entity should have only the `RoleName` attribute.

KevinKelly25:
Made corrections for comments brought up by @smurthys

Issue #141, opened 2018-01-04
The docs-development process may not support adding image and other files (WM)


smurthys:
**PS:** I am creating this issue so we can collectively discuss and receive notifications as new comments are added.

Context: We have add at least two instances where non-markdown files are to be included in the docs folder. See #136 and #139.

In general, we have the need to add HTML/Visio/image files to the docs folder with design notes and these files will have to be linked to from docs pages which are markdown (MD) files.

On the surface, this does not sound like much of an issue, but I worry our development process for docs may not support this ability: at least not very easily. Our docs-development process uses the Wiki as the "development environment" to write the docs and it worked very well in M1: We need to use the Wiki so anyone can easily update any page directly online and also avoid the approval process which can be both labor and time intensive.

It seems, if we want to continue to use the Wiki for doc development, we will need to clone the Wiki as a repository, add files to it, and check in the file. Even then, I am not sure if it will work correctly as evidenced in [@afig's Wiki page](https://github.com/DASSL/ClassDB/wiki/createRole()-User-Schema-checks-and-actions) on the flowchart related to Issue #136. See how the embedded image referencing a remote URL does not show up.

We need to **urgently** understand the underlying issues and learn workarounds/solutions so documentation does not become a problem. 

Perhaps there is no issue at all, in which case, it will be helpful to confirm so before we get too far and then find a problem.

KevinKelly25:
I took a look at [@afig's wiki page](https://github.com/DASSL/ClassDB/wiki/createRole()-User-Schema-checks-and-actions) and found that I could not use the URL link to the image that was embedded. I might be wrong, but there is a chance that it can just be google permissions not allowing the wiki page to access the image. 

As for embedding images using markdown I found [this](https://help.github.com/articles/adding-images-to-wikis/) help page which discusses adding images via markdown which might help.

So far, I have not found any work around for adding HTML files with MD. Most information I have seen about it suggests that MD only support a limited number of HTML tags. It may be that we can change ER-Schema to a MD file but wrap the code with ```. This won't display the HTML but until a work around is found it would allow all the files in the doc folder to be MD. 

I was not here for the M1 development process so I might misunderstanding the issue but these might be temporary fixes.

srrollo:
I'm testing embedding images that are stored locally in the repository, and it seems to work with both the wiki and GitHub pages. I'll post a link when I have a more complete example of what does/doesn't work. 

[...]

PR #153, opened 2018-01-09
Add PNG version of ER Schema

smurthys:
I created a PNG version of the ERD. 

The image is full size and is useful for viewing on its own. I also have a 80% size version for embedding, but we should verify if we can already resize an image by percentage (not absolute size) when embedded in an MD file. (This PR is related to Issue #141.)

A quick web search revealed these potentially useful links. I don't know if there is more current information:
- https://github.com/blog/1395-relative-links-in-markup-files
- https://webapps.stackexchange.com/questions/29602/markdown-to-insert-and-display-an-image-on-github-repo#29631
- https://stackoverflow.com/questions/24383700/resize-image-in-the-wiki-of-github-using-markdown#24384029


smurthys:
FYI, I used the Export function in draw.io. The key is to choose **border size of 10** instead of the default value of 0. I am documenting this information for when we need to recreate the image.

smurthys:
I like to leave this PR open until we have figured out if a second smaller embed-friendly image is also required.

srrollo:
I did a little playing around with image resizing, and I was unable to get and kind of % based resizing to work. It seems like only absolute resizing works on the wiki.

smurthys:
Commit 80e78b2 adds ER-Schema-small.png which shows the schema at 80% size. At 70%, the image was too small to read, but we can change it when we review the final docs.

smurthys:
Thanks @srrollo for verifying that sizes can't be specified in the Wiki.

Issue #156, opened 2018-01-11
The ER diagram is missing attributes (M)


smurthys:
The `User` entity in the ERD is missing the attribute `LastDDLOperation`.

Note: PNG versions need to be generated in 100% and 80% sizes for the updated ERD. See comments in PR #153 for additional information.

KevinKelly25:
Added PR #157 to address this issue in docs. However, I still need to modify the M2 design document in the wiki to reflect this missing attribute.

KevinKelly25:
Updated [M2 Design Outline](https://github.com/DASSL/ClassDB/wiki/User-and-Team-Management-(M2))

smurthys:
@KevinKelly25 I see the original error is in the design outline. Thanks for spotting and fixing that.

KevinKelly25:
Fixed by PR #157

PR #157, opened 2018-01-11
Update ERD

KevinKelly25:
Added LastDDLOperation to the ERD htlm, png, and small png for issue #156


KevinKelly25:
The updated PNGs look a more blurry then the originals. However, I am following the details left in PR #153. @smurthys other then exporting as png and having the border set to 10 was there any other options you had set? I am not sure what is causing the difference.

smurthys:
I used "File/Export as/Advanced" in draw.io and only set the border size to 10.

While you are there, please switch the order of `LastDDLObject` and `LastDDLOperation`.

Thank you for taking care of this issue.

KevinKelly25:
Fixed blurryness and switched order of LastDDLObject and LastDDLOperation.

For whatever reason. It seems if you have the same settings for File/Export as/png vs file/Export as/Advanced you will get different results, with the File/Export as/png coming out much more blurry.

smurthys:
The changes look good. Thanks @KevinKelly25 for fixing this issue.

Issue #146, opened 2018-01-06
New functions and views in Public are not removed by the uninstaller (WM)


srrollo:
[`addFrequentViews.sql`](https://github.com/DASSL/ClassDB/blob/add-frequent-views/src/addFrequentViews.sql) adds several new objects to `Public` that are not removed by `removeFromDB.sql`. I suggest that we simply add `DROP * IF EXISTS` statements to `removeFromDB.sql` for each object needing to be dropped. An initial list of such objects is:
- [ ]  `Public.describe(VARCHAR(63))`
- [ ] `Public.getMyActivitySummary()`
- [ ] `Public.getMyDDLActivity()`
- [ ] `Public.getMyConnectionActivity()`
- [ ] `Public.getMyActivity()`
- [ ] `Public.MyDDLActivity`
- [ ] `Public.MyConnectionActivity`
- [ ] `Public.MyActivity`
- [ ] `Public.MyActivitySummary`

smurthys:
I gathered some information about generating a script to list/drop functions. The internal function `oidvectortypes` is the magic sauce as seen in this query:

```sql
SELECT 'DROP FUNCTION ' || ns.nspname || '.' || p.proname 
       || '(' || oidvectortypes(p.proargtypes) || ');'
FROM pg_proc p JOIN pg_namespace ns ON p.pronamespace = ns.oid
WHERE ns.nspname = 'classdb'
ORDER BY p.proname;
```

Useful sources;
- https://stackoverflow.com/questions/10591113/drop-all-functions-from-postgres-database#10591166
- http://www.postgresonline.com/journal/archives/74-How-to-delete-many-functions.html
- https://www.postgresql.org/message-id/Pine.LNX.4.21.0007072058000.587-100000%40localhost.localdomain

srrollo:
This is an awesome function @smurthys. I was also thinking that somehow listing all functions owned by `ClassDB` would be the best way to solve this issue.

srrollo:
After investigating #147, it seems these two issues are linked fairly tightly. I will post an update here later.

srrollo:
I did some experimenting after our discussion, it seems one fairly simple solution we didn't discuss is to use `DROP OWNED BY ClassDB`. If this encounters any dependency issues, it throws a very helpful exception, for example:

`DETAIL:  view m2ins01.insview depends on table classdb.rolebase`

[...]

PR #161, opened 2018-01-12
Updated uninstall script for M2

### LINKED ISSUE #147: removeFromDB.sql relies on schema names matching user names (W)
`removeFromDB.sql` assumes that ClassDB user names match their schema names, and will always try and reassign the schema matching the user name instead of the user's actual schema. I believe getting the correct schema name with `ClassDB.getSchemaName()` should resolve this issue.

srrollo:
This PR updates the uninstall script to work with M2. I have changed to using the `DROP OWNED BY ClassDB` approach outlined [here](https://github.com/DASSL/ClassDB/issues/146#issuecomment-357379100). We can discuss if this solution is sufficient in our next sync, if necessary.

This method should no longer forcefully drop user objects that depend on ClassDB objects.

Fixes #147 and #146


srrollo:
Closing until dir. reorg is done

srrollo:
Reopened and editing PR comment

srrollo:
I just pushed another commit that sets `ClassDB.IDNameDomain`'s owner to `ClassDB`, while before it was owned by the superuser. This is necessary for `DROP OWNED BY ClassDB` to work, since many of the ClassDB functions depend on `ClassDB.IDNameDomain`.

When testing the uninstaller, I would recommend creating a new install of ClassDB from the scripts in this branch, as many missing assigns and grants have been added in dev. I ran into several problems running the uninstaller that were due to old functions that have since been removed from ClassDB, or incorrect function permissions that are now fixed.

smurthys:
Nice changes @srrollo. Using `DROP OWNED BY` is a good idea. (I should have suggested this approach earlier, because function `dropRole` already uses `DROP OWNED BY`.)

I have the following thoughts on `removeAllFromDB.sql`:
- The file prologue should probably say that the script stops if drop fails at any point (due to not using DROP...CASCADE)
- L101: The only reason to have this line is if the `ClassDB` schema is somehow not owned by the `ClassDB` role (and thus the schema is not already dropped by L98). Perhaps a comment to that effect helps.
- I recommend moving L98 and L101 to immediately after L 84.
- Doesn't the comment on L90 in general also apply to code in L87-L88?

Also, thank you for adding ownership to `ClassDB.IDNameDomain`. I'd missed doing that.

smurthys:
FY, I'll be away for 60-90 minutes. So, please feel free to merge this PR unless one of you discovers something major. I'll approve the PR in anticipation of there being no such problem.

afig:
Thanks for working on the uninstall script @srrollo. I tried running it on a clean M2 install, but received the following error (showing full output since it may be helpful in determining where the script fails):

```
classdbtest=# \i removeAllFromDB.sql
START TRANSACTION
SET
DO
DO
DO
DROP OWNED
DROP OWNED
DROP OWNED
DROP EVENT TRIGGER
DROP EVENT TRIGGER
DROP FUNCTION
DROP FUNCTION
DROP FUNCTION
DROP FUNCTION
psql:removeAllFRomDB.sql:98: ERROR:  cannot drop desired object(s) because other objects depend on them
DETAIL:  event trigger triggerddlcommandsqldrop depends on function classdb.logddlactivity()
```

Should `DROP EVENT TRIGGER IF EXISTS triggerDDLCommandSqlDrop;` be added to the uninstall? Likewise, should the accompanying `triggerDDLCommandEnd` also be added?

srrollo:
Thanks for the review @smurthys. Thanks for pointing this out @afig. I had somehow pasted the wrong names for the event triggers in the `DROP EVENT TRIGGER` statements. The latest commit should fix this.

Issue #142, opened 2018-01-05
addConnectionMgmt.sql is incorrectly included in the core installation scripts (E)


smurthys:
The script `addConnectionMgmt.sql` is included in `prepareDB.psql` though the functionality `addConnectionMgmt.sql` adds is not essential to using ClassDB: the functionality added is useful, but not essential.

I propose excluding `addConnectionMgmt.sql` from `prepardDB.psql`.

smurthys:
I also propose the following improvements :
- Rename `prepareDB.psql` to `prepareDBCore.sql`
- Create a `prepareDBAddOn.psql` with entries for scripts for each optional add-on feature.

Proposed organization for `prepareDBAddOn.psql`:
- Group script files by feature and dependency, with a blank line before and after each group, and a meaningful comment before the start of the group 
- As of now, most features need only one script file

smurthys:
Also, we should classify each script file as "Core" or "AddOn" and add the file to appropriate `.psql` file

smurthys:
We made the following the decision in the Skype call on Jan. 10, 2017 about this topic:
- Scripts are of three kinds: core, recommended, and optional
- Every script is of exactly one kind
- Each kind of script will be stored in its own kind-specific sub-directory (of directory `src`). The sub-directory names are: `core`, `reco`, and `opt`
- Each script's filename has a suffix to denote its kind. Though they may seem redundant, the suffixes make it easier to spot and discuss files. The suffixes are `Core`, `Reco`, and `Opt`. For example, `addRoleBaseMgmtCore.sql`.

In addition to the structure summarized above, I propose server-level scripts and db-level scripts be stored in separate sub-directories, and each of them further containing kind-specific sub-directories. A kind-specific sub-directory may be not created at all if no script of that kind is present.

The following is the proposed directory structure for source files:

```
src
  -- server
    -- core
    -- reco
    -- opt
  -- db
    -- core
    -- reco
    -- opt
```

afig:
Resolved by #164

PR #164, opened 2018-01-13
Source directory reorganization

afig:
This PR separates all script files into one of three kinds: core, recommended, and optional (`core`, `reco`, and `opt` directories, respectively). Additionally, script files are now also separated into database scripts and server scripts (`db` and `server` directories under `src`, respectively).

Suffixes were added to file names corresponding to which of the three kinds they belong to.

For more information, see issue #142, which this PR fixes.

In terms of review, ensure that files are properly categorized. Particularly `addCatalogMgmtOpt.sql`, as I do not recall whether it was decided if it was recommended or optional.

Also, I left the uninstallation scripts (`removeFromDB.sql` and `removeFromServer.sql`) uncategorized (outside of a `core`, `reco`, or `opt` directory), as they remove all portions of the system.

I updated the file names inside of the `prepareDBCore.psql` script, added `addUserMgmtCore.sql` and `addClassDBRoleViewsCore.sql`, and removed `addConnectionMgmt.sql`.


srrollo:
Looks good @afig. I think all the scripts have the correct placement. IMO, based on our discussion, `addCatalogMgmtOpt.sql` should be Opt.

The script names still need to be changed inside each file, but I imagine this is best left until we are sure of the location of each script. Also, should we include the 'full' path (ie. `db/core/addXYZCore.sql`), or just the filename in the prologue comments?

afig:
Thanks for pointing out that the file names have to be updated for the headers, I had overlooked that when making these changes. However, I think it might over-complicate things if we add the full path.

I agree that the currently placement of `disableConnectionLoggingReco.psql` is rather confusing, but then again, it's not really an "uninstall" script either... 😕. I think it will be best to leave it for now and discuss it at some other point.

smurthys:
Excellent work and analysis @afig and @srrollo.

I like the current location of the `remove*.sql` files. I recommend renaming them to `removeAllFrom*.sql` so it is obvious they will remove everything.

smurthys:
I recommend renaming the `removeFrom*.sql` files as part of this PR to avoid issue-management effort. However, I am OK if the renaming is done after due to other considerations.

afig:
The two files were renamed and headers updated:

`removeFromDB.sql` -> `removeAllFromDB.sql`

`removeFromServer.sql` -> `removeAllFromServer.sql`

smurthys:
Nicely done. I'm eager to see how the new directory structure plays out.

PR #167, opened 2018-01-13
Add kind specific prep scripts

KevinKelly25:
Added kind specific prep scripts for each subfolder in server and database.

### DIFF
--- a/src/db/core/addAllDBCore.psql
+++ b/src/db/core/addAllDBCore.psql
@@ -22,10 +22,13 @@
 --  the script can be run using the -f command line switch of psql or using the \i command
 --  inside a psql interactive session
 
+--This script uses the \ir flag so that a relative path from where the script is located is
+-- used rather then the program's current directory
+
 \set ON_ERROR_STOP on
-\i initializeDBCore.sql
-\i addHelpersCore.sql
-\i addRoleBaseMgmtCore.sql
-\i addClassDBRolesMgmtCore.sql
-\i addUserMgmtCore.sql
-\i addClassDBRolesViewsCore.sql
+\ir initializeDBCore.sql
+\ir addHelpersCore.sql
+\ir addRoleBaseMgmtCore.sql
+\ir addClassDBRolesMgmtCore.sql
+\ir addUserMgmtCore.sql
+\ir addClassDBRolesViewsCore.sql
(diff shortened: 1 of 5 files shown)


smurthys:
Thanks @KevinKelly25 for creating the files. The files generally implement the criteria we decided.

Before it is lost in other details, I wonder if `prepareServerCore.sql` should be renamed to `intitializeServerCore.sql`. That is not the subject of this PR, but this PR does show the need to rename. If we agree, the change can be done in this PR.

A few thoughts:
- In all but `addAllDBCore.psql`, please add your name to the end of the attribution list
- `addAllServerReco.psql` should contain only `enableConnectionLoggingReco.sql`. @srrollo can confirm 
if this method of enabling logging works. (I believe it does.)
- In all scripts, I recommend splitting comments such as the one that begin on L18 in `addAllDBOpt.psql` as follows (with suitable modifications to each file). BTW, `addAllServerReco.psql` does not depend on anything.

```
--This script runs other scripts that add optional ClassDB functionality to the current database

--This script should be run after adding core ClassDB functionality
 
-- This script uses psql meta-commands run other scripts and so it should be run using psql.
--  the script can be run using the -f command line switch of psql or using the \i command 
--  inside a psql interactive session
```

KevinKelly25:
Made corrections for @smurthys  comment

afig:
Thanks for implementing these scripts @KevinKelly25, they work great.

In relation to @smurthys's comment, I agree that `prepareServerCore.psql` should be renamed to `initializeServerCore.psql`.

On a related note, I wonder if we should [use the `\ir` flag](https://www.postgresql.org/docs/9.6/static/app-psql.html) instead of the `\i` flag that we are using now. The `\i` flag looks for scripts using the current working directory of the command window. However, `\ir` looks for scripts using a relative path from where the script is located. Using this new flag will avoid having to exit `psql` and change directories to run each of the new scripts. 

- To save time and sanity, this flag change does not have to be done manually; the "find and replace in files" feature of a text editor will make this process a relatively simple task.

- The `\ir` flag was introduced in the version of `psql` bundled with Postgres 9.2

(Just got ninja'd by @srrollo 😄 )

smurthys:
I agree with using `\ir` meta command.

srrollo:
I actually agree with @afig's assesment more than mine. I though it would be enough to execute the install scripts with the `\ir` command, but it appears we need to change `\i` in all the scripts for it to work.

smurthys:
BTW, comments in scripts should also reflect the use of `\ir` meta command.

KevinKelly25:
@afig and @srrollo Thank you for making me aware of the \ir flag. It makes much more sense to use. I added commit c897652 to address that issue and add related comments.

PR #172, opened 2018-01-14
Fix user view

### LINKED ISSUE #169: User view has NULL value for DDLCount and ConnectionCount (W)
Attributes `DDLCount` and `ConnectionCount` are `NULL` instead of zero for users without DDL activity and connection activity respectively.

### LINKED ISSUE #171: Field HasClassDBRole can return TRUE even if user does not have a ClassDB role (W)
Field [`HasClassDBRole`](https://github.com/DASSL/ClassDB/blob/d2cdb4666a3d12ec42b9caadbd9f26334d4143e3/src/db/core/addClassDBRolesViewsCore.sql#L48) returns whenever a role is recorded as a user even if the user is not a member of one of the three ClassDB roles expected (instructor, student, dbmanager)

smurthys:
Commit e673c33 fixes #169. 

Commit 6ac9b41 fixes #171. It also:
- Changes how last DDL op and object are retrieved: it no longer user a derived table.
- Re-formats code for line length
- Updates comments

### DIFF
--- a/src/db/core/addClassDBRolesViewsCore.sql
+++ b/src/db/core/addClassDBRolesViewsCore.sql
@@ -40,35 +40,47 @@ $$;
 
 
 --Define a view to return known users
+-- IsXYZ fields denote if user is a member of the appropriate ClassDB role
+-- HasClassDBRole is TRUE if any of the IsXYZ fields is TRUE
+-- creates derived tables w/ user-specific aggregates over tables DDLActivity
+-- and ConnectionActivity
+-- DISTINCT ON (RoleName) limits output to just one row per role: required
+-- because the join to get last DDL op and object can technically return
+-- more than one row for a user because DDLActivity does not have a PK
 CREATE OR REPLACE VIEW ClassDB.User AS
-  SELECT RoleName AS UserName, FullName, SchemaName, ExtraInfo,
-  ClassDB.IsInstructor(RoleName) AS IsInstructor, --True if user is instructor
-  ClassDB.IsStudent(RoleName) AS IsStudent, --True if user is student
-  ClassDB.IsDBManager(RoleName) AS IsDBManager, --True if user is DBManager
-  ClassDB.isUser(RoleName) AS HasClassDBRole, --True if user is any ClassDB role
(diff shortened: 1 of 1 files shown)


smurthys:
Here is a quick test I used (there isn't a proper unit test for this script in the repo):

```sql
--create 4 users: students s1, s2; instr i1: dbman d1
SELECT ClassDB.createStudent('s1', 's1 name');
SELECT ClassDB.createStudent('s2', 's2 name');
SELECT ClassDB.createInstructor('i1', 'i1 name');
SELECT ClassDB.createDBManager('d1', 'd1 name');

--create a team t1: should not be in User view
SELECT ClassDB.createRole('t1', 't1 name', TRUE);

--clear the activity tables
TRUNCATE ClassDB.ConnectionActivity;
TRUNCATE ClassDB.DDLActivity;

--add connection activities: 1 for s1, none for s2 and d1, 2 for i1
INSERT INTO ClassDB.ConnectionActivity VALUES
('s1', TIMESTAMP '2018-01-01 10:23:54'),
('i1', TIMESTAMP '2018-01-01 01:10:00'),
('i1', TIMESTAMP '2018-01-05 13:41:56');

--add ddl activies:
-- 4 for s1: two of them with the same timestamp
-- 1 for s2 (not ordinarily possible if s2 has no connections)
-- none for i1 and d1
INSERT INTO ClassDB.DDLActivity VALUES
('s1', TIMESTAMP '2018-01-01 10:25:13', 'CREATE TABLE', 'public.x'),
('s1', TIMESTAMP '2018-01-01 10:32:16', 'DROP TABLE', 'public.x'),
('s1', TIMESTAMP '2018-01-01 10:46:54', 'CREATE SCHEMA', 'why_not'),
('s1', TIMESTAMP '2018-01-01 10:46:54', 'DROP TABLE', 'dupe_activity'),
('s2', TIMESTAMP '2018-01-04 01:22:00', 'CREATE VIEW', 'public.v');

SELECT * FROM ClassDB.User;
SELECT * FROM ClassDB.Instructor;
SELECT * FROM ClassDB.Student;
SELECT * FROM ClassDB.DBManager;
```

PR #174, opened 2018-01-14
Fixed typo in server reco

### LINKED ISSUE #173: addAllServerReco.psql fails due to a type (W)
`addAllServerReco.psql` tries to execute `enableConnectionLoggingReco.sql` instead of `enableConnectionLoggingReco.psql`.

srrollo:
See #173 - Fixes #173


smurthys:
Glad you caught this issue. It is amazing how many little things can go wrong no matter how careful we all are. 😢

PR #178, opened 2018-01-14
Update privilege tests

afig:
Previously, none of the privilege tests functioned as they should since they were still using V1.0's API and objects. They have now been updated to work with M2 (V2.0), and have had new objects added to which privileges needed to be tested.

No issues have been identified with the current state of privileges. Also, no major issues were identified with the functionality of the system.

Additionally, running privilege tests no longer require the user to manually reconnect to the database and type in a script name. This is now taken care of by a `psql` script. The readme has also been updated to reflect this.

Many code-consistency fixes were also make to the tests themselves.

No modifications to the system were done in this PR.

### DIFF
--- a/tests/privileges/0_setup.sql
+++ b/tests/privileges/0_setup.sql
@@ -1,7 +1,8 @@
 --0_setup.sql - ClassDB
 
 --Andrew Figueroa, Steven Rollo, Sean Murthy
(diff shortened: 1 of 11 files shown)


afig:
Thanks for the late night reviews @smurthys and @srrollo . Also wanted to thank @srrollo for using the `\c` command and storing `postgres` in a variable in another script, since that's what led me to create the `runAllPrivilegeTests.psql` script.

Issue #182, opened 2018-01-14
Views StudentTable and StudentTableCount can return information unrelated to students (W)


smurthys:
The problem is that `getSchemaName` is only reliable in the transaction where the user is created.

The views can simply consider all tables owned by each student without regard for schema name, and include both user name and schema name in the result.

This issue is related to Issue #181,

PR #183, opened 2018-01-14
Fix StudentTable and StudentTableCount

### LINKED ISSUE #181: Views StudentTable and StudentTableCount do not contain user name (M)
Views `StudentTable` and `StudentTableCount` should include user name.

EDIT: It presently includes schema name possibly with the assumption that schema name is same as owner name, which is not true.

smurthys:
Commit 9765503 fixes #181 and #182
- View `StudentTable` now returns both user name and schema name, and it works on all tables and views students own independent of schema name and ownership.
  - It now uses `pg_catalog` instead of `INFORMATION_SCHEMA`, because the latter does not provide all the information needed
- View `StudentTableCount` now returns user name and just reuses view `StudentTable`
- Columns in both views now have appropriate names
- General: adjusted line spaces and put ownership grants first in permission blocks.

### DIFF
--- a/src/db/reco/addFrequentViewsReco.sql
+++ b/src/db/reco/addFrequentViewsReco.sql
@@ -19,7 +19,7 @@
 --This script creates several objects (we collectively refer to them as views) to
 -- display summary data related to student activity in the current database.
 -- Views that are accessible to students and require access to ClassDB.User are
--- implememted as functions. This allows the views to access the ClassDB schema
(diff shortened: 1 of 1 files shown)


smurthys:
Thanks @srrollo and @afig for the review.

Commit 7e9712a gives the `count(*)` column a name in view `StudentTableCount`.

BTW, the file in question has a few minor issues with line lengths, etc., but those things will have to wait.

Issue #136, opened 2017-12-28
Missing checks if user/schema already exists (M)


smurthys:
[Function createUser in addUserMgmt.sql](https://github.com/DASSL/ClassDB/blob/05872e166a85db91d76e3d620d0c1b0ba20229ec/src/addUserMgmt.sql#L44-L64) does not seem to handle a few issues if a user/schema already exists

* User already exists causes a notice: does not (and cannot) check if the user who already exists is the same as the user being created.
  - The only way to check "user sameness" would be to compare e-mail address but this function does not have access to that information. See Issue #49.

* Schema already exists causes a notice, but does not check if the user has permission to that schema

This issue is created based on the discussion of Issue #1.

afig:
Raising an notice/exception would be the best course of action. Given that other parts of ClassDB depend on there being a correspondence between user and schema names, I believe that an exception would be best, so that there doesn't exist the possibility of there being a user without a corresponding schema they have access to. At minimum, a warning should be raised

It might be helpful to view the table of [PostgreSQL's message severity levels](https://www.postgresql.org/docs/current/static/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS), copied here:

| Severity | Usage | syslog | eventlog |
|----------------|------------------------------------------------------------------------------------------------------|---------|-------------|
| DEBUG1..DEBUG5 | Provides successively-more-detailed information for use by developers. | DEBUG | INFORMATION |
| INFO | Provides information implicitly requested by the user, e.g., output from VACUUM VERBOSE. | INFO | INFORMATION |
| NOTICE | Provides information that might be helpful to users, e.g., notice of truncation of long identifiers. | NOTICE | INFORMATION |
| WARNING | Provides warnings of likely problems, e.g., COMMIT outside a transaction block. | NOTICE | WARNING |
| ERROR | Reports an error that caused the current command to abort. | WARNING | ERROR |
| LOG | Reports information of interest to administrators, e.g., checkpoint activity. | INFO | INFORMATION |
| FATAL | Reports an error that caused the current session to abort. | ERR | ERROR |
| PANIC | Reports an error that caused all database sessions to abort. | CRIT | ERROR |

A message with a severity of `ERROR` would be raised if we decide to go with an exception.

smurthys:
I propose starting a Wiki page with a flow chart showing the conditions and actions. Though it might sound like a lot of work, in all my experience, this approach will likely take less time overall: reduces number of code iterations.
 
[...]

PR #189, opened 2018-01-15
Add HTML and PNG versions of createRole flowchart

afig:
This flowchart describes the sequence in which the `createRole()` function located in `addRoleBaseMgmtCore.sql` performs its checks and what corresponding actions are taken.

For more information, see issue #136


smurthys:
Thanks for adding the files @afig.

I think the following improvements are possible:
- Change HTML title to something like `ClassDB Role-Creation Process`
- Add a heading to the chart with the same text as the document title. This will be similar to what is done in the ERD.
- I like the idea of the current text showing the signature of createRole. Perhaps that could be shown as the sub-heading in the same font size as now but saying `Function createRole(roleName,...initialPwd)`
- One thing in ERD that is very helpful is to show somewhere that the chart is drawn using draw.io. I don't quite know where that fits.

afig:
Thanks for the recommendations @smurthys. I took them into account and updated the title, and added one within the diagram. I also tweaked the text that showed the signature of `createRole`.

I split the last line of the attribution box in order to reduce the box's overall width. I was then able to add a draw.io mention at the bottom of the page without it sticking out too much.

smurthys:
I think these changes will do.

PR #193, opened 2018-01-17
Added new install all scripts

### LINKED ISSUE #185: Some installation files are missing or have incorrect names (WM)
There are no installation scripts for installing all components at once, and one installation script has an incorrect name.
A proposed solution is:
1. [x] `prepareServerCore.sql` => `initializeServerCore.sql`
2. [x] Add `addAllDB.psql`, which installs all DB components
3. [x] Add `addAllServer.psql` which installs all server components

The two new installer scripts should call the existing `addAllXY.psql` scripts instead of individual sql files.

srrollo:
This PR adds the new install scripts and renames existing ones as discussed in #185. 

Fixes #185


srrollo:
I've just pushed two commits that update the file names based on our latest decision in #185. Additionally, I've removed the extra space present in the comment at @afig pointed out in all `.psql` install scripts.

PR #196, opened 2018-01-20
Make logging tests use new drop dispositions and improve user views

srrollo:
This is a collection of small fixes/improvements to the logging tests scripts and user views. I intended for this to be two separate PRs, but I accidentally merged all the changes into one commit.

- The logging tests no longer fail due to the new drop dispositions. The new [removing users](https://github.com/DASSL/ClassDB/wiki/Removing-Users) page helped a lot with making this change easy (@afig)
- The `ClassDB.StudentActivitySummaryAnon` view now calls the correct underlying function
- `ClassDB.getUserAcitivity()` orders activity records by username, so all views/functions that return individual activity records are now ordered by username.

With these changes, all tests pass a new ClassDB instance setup using the new quick-start guide 🎉

### DIFF
--- a/src/db/core/addClassDBRolesViewsCore.sql
+++ b/src/db/core/addClassDBRolesViewsCore.sql
@@ -79,7 +79,8 @@ LEFT OUTER JOIN
   FROM ClassDB.ConnectionActivity
   GROUP BY UserName
 ) AS C ON C.UserName = RoleName
-WHERE NOT IsTeam;
+WHERE NOT IsTeam
+ORDER BY UserName;
 
 
 ALTER VIEW ClassDB.User OWNER TO ClassDB;
--- a/src/db/reco/addFrequentViewsReco.sql
+++ b/src/db/reco/addFrequentViewsReco.sql
@@ -191,7 +191,7 @@ CREATE OR REPLACE VIEW ClassDB.StudentActivitySummaryAnon AS
 (
    SELECT DDLCount, LastDDLOperation, LastDDLObject, LastDDLActivityAt,
           ConnectionCount, LastConnectionAt
-   FROM   ClassDB.getStudentActivitySummary()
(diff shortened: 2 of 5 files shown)


smurthys:
Thanks @srrollo for the changes. I totally agree the doc on removing users is extremely helpful.

I have the following observations in `addFrequentViewsCore.sql` (some of these I find as I review the fie):
- `StudentTable`: needs `ORDER BY UserName, SchemaName, TableName`
- `getUserDDLActivity`: needs `ORDER BY UserName, StatementStartedAt DESC`
- `getUserConnectionActivity`: needs `ORDER BY UserName, AcceptedAt DESC`
- `getUserActivity`: needs `ORDER BY UserName, ActivityAt DESC` (see below)
  - Though not required, I recommend renaming column `StatementStartedAt` as `ActivityAt` in the first query so it is easy to relate the columns in `ORDER BY` to the final output

Related, I also suggest adding `ORDER BY UserName` to view `User` in `addClassDBRolesViewsCore.sql`. This change will automatically impose ordering on the result of some functions in `addFrequentViewsCore.sql`.
- There is no guarantee that ordering percolates from a table/view to a derived table/view, but this is the smallest change we can make now.

I have not had a chance to look at the test scripts, but I notice a `COMMIT` has changed to `ROLLBACK`. This is the correct approach. However, in a test script that uses `ROLLBACK`, it is not necessary to drop/delete objects unless drop/delete is part of the test.

I just created a new issue #197 about using `ROLLBACK` instead of `COMMIT` in test scripts, but we don't need to address that issue in M2.

srrollo:
Thanks for the review @smurthys. I've made the suggested changes to ordering. 

Good point about not dropping objects in tests when not using `ROLLBACK`. In this case, `testAddConnectionActivityLogging.psql` can't use `ROLLBACK`, because the test spans multiple transactions due to changing connections.

Issue #200, opened 2018-03-29
Error importing connection log (W)


smurthys:
Executing  `select * from classdb.importconnectionlog();` on the CS205 DB in production failed, and produced the following error message:

BTW, the IP address of the host has changed over time, but that should not cause the error reported.

```
ERROR:  could not open file "/opt/postgresql/data/logs/pg96/postgresql-03.04.csv" for reading: No such file or directory
CONTEXT:  SQL statement "COPY classdb.postgresLog FROM '/opt/postgresql/data/logs/pg96/postgresql-03.04.csv' WITH csv"
```
![image](https://user-images.githubusercontent.com/29356581/38097560-b563f4be-3343-11e8-9c89-e9568e22e603.png)

smurthys:
IMPORTANT: For security sake, please be careful including details of the server or the DB in this thread, because the issue is about a system in production.

KevinKelly25:
I am not sure exactly what happened with the server earlier this semester but when the server went down could the CSV file have gotten deleted, renamed, or the path changed?

srrollo:
This issue should probably be kept open - `classdb.importconnectionlog()` should have some sort of error handling added to deal with this situation.

srrollo:
I was able to log onto the server and confirm that only the 03-04 log file is missing. This is causing the import error. The solution outlined in the above comments still holds.

srrollo:
The issue on the local server has been fixed by "patching" the log files.  
For future reference, I have two general ideas for approaching the code patch:    

- Check if log files exist before trying to import
- Use exception handling to catch the error when importing a missing file

The first is not possible without using a scripting language function, I need to do more research on the second to see if the file read error is handleable.

srrollo:
PL\pgSQL does support exception handling (see [here](https://www.postgresql.org/docs/10/static/plpgsql-control-structures.html#PLPGSQL-ERROR-TRAPPING) and [here](https://www.postgresql.org/docs/10/static/errcodes-appendix.html)).  
I have used this to add some simple exception handling to `ClassDB.importConnectionLog()` that solves the issue described in this thread by skipping missing log files. See [here](https://github.com/DASSL/ClassDB/compare/log-exception-handling) for a diff. I think there are a few points to consider with how missing log files should be handled:
- Should they be skipped silently or display a message
- Should `ClassDB.importConnectionLog()` still fail but with a more descriptive error message
- Should there be a manual override parameter to skip missing log files

Any input would be appreciated.

[...]

PR #203, opened 2018-06-01
Enable logging collector

### LINKED ISSUE #202: `enableConnectionLoggingReco.psql` assumes `logging_collector` is set to `on` (W)
The Postgres distribution in the Ubuntu repository (which most Ubuntu users are likely to run) turns off the logging collector by default. If the logging collector is off, all log data is sent to `stderr` (which gets picked up by systemd), regardless of any other log settings. `enableConnectionLoggingReco.psql` should set `logging_collector` to `on`, which ensures log data will be sent to the expected file.

srrollo:
This PR fixes #202 by enabling `logging_collector` in `enableConnectionLoggingReco.psql`. I am creating this PR so others can test the code if they wish. I believe approval should be withheld for now, because this change will require some significant documentation changes:

Unfortunately, `logging_collector` is fixed at runtime, so now users will have to restart their Postgres instance in order to enable connection logging. My concern is that restarting the instance introduces a significant roadblock to anyone installing ClassDB to an in-production system. I am debating whether the documentation should unconditionally state that users need to restart that their Postgres instance, or if the script should check if `logging_collector` is already enabled and notify the user that they do not need to restart their instance. Any input is welcome.


srrollo:
I also think that adding the check is the best course of action. I think we also need to highlight that `ClassDB.importConnectionLog()` won't work until after the restart. I think this will have to work with the solution to #200, since that will handle missing log files more gracefully. Perhaps `ClassDB.importConnectionLog()` can even check `logging_collector` when it starts and provide a helpful error message.

smurthys:
Thanks for identifying the potential issue and making the change @srrollo.

* I wonder if the order of turning on `log_connections` and `logging_collector` matter. It probably does not
* I recommend conditionally raising a warning that a restart is needed if we just turned on `logging_collector`. I recommend against raising notice because users habitually disable/ignore them

srrollo:
I pushed another commit that raises a warning if `logging_collector` is changed from `off` to `on`. I used an anon code block for this, since there is no guarantee any ClassDB functions will exist when the server scripts are run. I also reordered the `ALTER SYSTEM` statements to more logically show the changes made to the Postgres config.

smurthys:
I have two sets of recommendations. 

At start of script:
- Assert current user is superuser: just copy code from `addHelpersCore.sql`
- Use [SELECT ... INTO](https://www.postgresql.org/docs/10/static/plpgsql-statements.html) to store the current value of `logging_connector` in a variable.
- Use `COALESCE` or other means to store `FALSE` in the variable to handle the `WHERE` predicate failing: I don't know what Postgres does, but DBMSs tend to not store a configuration row if a value has not been explicitly set. (An alternative is to use `IF NOT FOUND` immediately after `SELECT`.)
- Note: `NOT(SELECT x ...)` is "unknown" is `SELECT` returns no rows. As written, this situation results in no warning to be raised

Structure the code as follows:
- Store the current value of `logging_connector` in a variable
- Execute `ALTER SYSTEM` as is presently done
- Reload configuration if `logging_connector` was already on; else raise a warning about restarting the server

afig:
About @smurthys's comment:

- Interesting point about the ```WHERE``` predicate failing. The behavior noted about not having rows for unset parameters doesn't seem to apply to ```pg_settings``` (possibly because it is a view and already performs an internal ```COALESCE```), however, it is certainly good practice to perform the above behavior and worth keeping in mind for other scenarios.

- Unfortunately, we cannot make a full assertion since as mentioned in L23-L25, these ```ALTER SYSTEM``` commands cannot be run in a transaction block. We can raise a message, but the remaining commands will still attempt to execute (and will all raise their own message about not being a ```superuser```). However, it may still be worthwhile to raise our own message.

***

Apart from the changes already suggested, this looks good. I tested this on an instance that had the ```logging_collector``` turned off and the script successfully turned it on while issuing a ```WARNING```. It also did not raise a message if the ```logging_collector``` was already turned on. This will need to be reevaluated after any changes, however.

I think having ```ClassDB.importConnectionLog()``` check for the state of the ```logging_collector``` is somewhat unnecessary. Rather, this may be better addressed by expanding the [Troubleshooting page](https://dassl.github.io/ClassDB/Troubleshooting) on the ClassDB docs to include issues that may come up when importing logs. 

We should also update the [Setup page](https://github.com/DASSL/ClassDB/wiki/Setup) to discuss the potential need for a restart.

smurthys:
Thanks @afig for the reminder about the per-line execution nature of this script. I totally forgot that aspect.

BTW, I have some observations on logging related checks in `ClassDB.importConnectionLog()`, but have not shared them here because that code is (presently) not part of this PR.

srrollo:
Thanks for the feedback. I suspect that using `SELECT INTO` will not work in this situation either, because it is a PL/pgSQL statement. Since all `BEGIN END` blocks are atomic, I suspect we will be unable to use `ALTER SYSTEM` within them. I will verify this behavior later.

If this proves to be the case, perhaps we can place the `IF` statement after the `ALTER SYSTEM` statements that either executes `pg_reload_conf()` or notifies the user they need to restart.

srrollo:
Also, I plan on introducing other logging changes in a separate PR. I think they still need some work, but I can open that PR so we can begin a discussion.

smurthys:
I am curious about the choice to [turn on logging_collector at end](https://github.com/DASSL/ClassDB/blob/088148622960c0216385a57aa11558c28251a531/src/server/reco/enableConnectionLoggingReco.psql#L61). Shouldn't that be done before reloading the changed config?

srrollo:
Technically, `pg_reload_conf()` does not need to be called when `logging_collector` is set, because the server must be restarted for the setting to take effect. Once the server is restarted, all new settings value are used, even if `pg_reload_conf()` was never called. That being said, I had arranged the `ALTER SYSTEM` statements in this way to get around the perceived limitation that we would need to store the value of `logging_collector`. 

After thinking about this a bit, I realized that `logging_collector` always reflects the current effective value, so setting `logging_collector` before the `IF` statement does not matter. I initially rewrote this to store `logging_collector` in a temporary table, which works but is not really necessary. If we feel that storing the original value is safer, I can re-add that code. 

@afig Thanks for linking the exception style guide. It's pretty helpful that they actually documented this. I've updated the error message based on your suggestion.

For reference, here is the temp table code:
```sql
CREATE TEMPORARY TABLE LoggingCollectorValue AS
SELECT COALESCE(setting::BOOLEAN, FALSE) AS value
FROM pg_settings
WHERE name = 'logging_collector';
```

smurthys:
Actually, I had written the wrong setting name. I meant to say [`log_connections`](https://github.com/DASSL/ClassDB/blob/088148622960c0216385a57aa11558c28251a531/src/server/reco/enableConnectionLoggingReco.psql#L61), which indeed must be set before reloading config. However, I see the newest commit does this correctly.

IMO, [`logging_collector`](https://github.com/DASSL/ClassDB/blob/ced886ba42b9db5c13c167eb5cfab2c5d5c4bb75/src/server/reco/enableConnectionLoggingReco.psql#L37) should be conditionally turned ON just before raising warning. 

The only reason to turn on `logging_collector` early is if config requires it in that order.

srrollo:
The order does not matter. I have placed `logging_collector` at the end in the latest commit.

smurthys:
I agree `logging_collector` can be forcibly turned on before checking if it is on, but that code does not flow well and is not as maintainable as turning that parameter on conditionally:

The following organization is more natural:
```
set parameters other than logging_collector
if logging_collector is off
   set logging_collector on
   raise warning
else
  reload configuration
```

afig:
Thanks for taking the time to update the docs @srrollo. I agree that the organization pointed out by @smurthys in the previous comment is preferred, however, it cannot be easily implemented because setting `logging_collector` to `on` uses the `ALTER SYSTEM` command, which cannot be performed inside of a transaction block.

Apart from this remaining discussion, I think this looks good to merge.

smurthys:
Don't mind this absent-minded professor. At least in this thread. I'll strive to be more conscious.

Issue #208, opened 2018-06-05
Function ClassDB.isStudent has unnecessary code (E)


smurthys:
Function `ClassDB.isStudent` (and 2 other functions) has the [following unnecessary code](https://github.com/DASSL/ClassDB/blob/14eb33909f670c0689a791e32dbd4bd3cd596784/src/db/core/addClassDBRolesMgmtCore.sql#L50-L56) which can be simplified (a possible rewrite is also shown):

```sql
IF NOT ClassDB.isUser($1) THEN
   RETURN FALSE;
ELSIF NOT ClassDB.isMember($1, 'classdb_student') THEN
   RETURN FALSE;
ELSE
   RETURN TRUE;
END IF;
```
**The code can be simplified as follows:**

```sql
IF NOT ClassDB.isUser($1) THEN
   RETURN FALSE;
ELSE
   RETURN ClassDB.isMember($1, 'classdb_student');
END IF;
```

PR #209, opened 2018-06-06
Remove unnecessary code from isXYZ functions

afig:
Simplifies code in ```isStudent```, ```isInstructor```, and ```isDBManager``` functions within ```addClassDBRolesMgmtCore.sql``` as pointed out in issue #208. All unit test continue to pass after this (small) change.


srrollo:
Changes look good, all tests pass. Should be good to merge

Issue #201, opened 2018-05-30
createXYZ functions accept a potentially unsafe parameter (E)


smurthys:
The parameter `initialPwd` in functions[ `ClassDB.createStudent`](https://github.com/DASSL/ClassDB/blob/14eb33909f670c0689a791e32dbd4bd3cd596784/src/db/core/addClassDBRolesMgmtCore.sql#L125-L132), `ClassDB.createInstructor`, and `ClassDB.createDBManager` is transmitted in clear if the connection used to invoke these functions is not secure.

The potential loss of security is not really an error in ClassDB, but rather of the user. However, in the past, we have removed the function `changePassword` for the reason that we want to remove any scope for insecure practices.

I propose the following changes:
- Add a new version of the functions without the pwd parameter and use the default pwd policy currently in place
- Alter the existing functions to ignore any pwd passed. Simply call the new functions and also raise a warning that the function is deprecated.

smurthys:
The [code that needs to be fixed](https://github.com/DASSL/ClassDB/blob/14eb33909f670c0689a791e32dbd4bd3cd596784/src/db/core/addClassDBRolesMgmtCore.sql#L125-L132) is as follows. Because parameter `initialPwd` has a default value, I believe it is not possible to define an overload without just that parameter. I will propose a workaround in the next comment.

```sql
ClassDB.createStudent
(userName ClassDB.IDNameDomain,
 fullName ClassDB.RoleBase.FullName%Type,
 schemaName ClassDB.IDNameDomain DEFAULT NULL,
 extraInfo ClassDB.RoleBase.ExtraInfo%Type DEFAULT NULL,
 okIfRoleExists BOOLEAN DEFAULT TRUE,
 okIfSchemaExists BOOLEAN DEFAULT TRUE,
 initialPwd VARCHAR(128) DEFAULT NULL
)
```

smurthys:
The ideal solution is to define an overload function without the `initialPwd` parameter (the last parameter), but such an overload should result in an error at function call due to the ambiguity the overload introduces.

However, unlike C++, it seems Postgres [overload rules](https://www.postgresql.org/docs/current/static/typeconv-func.htmlrl) permit overloads that normally result in ambiguous function calls, by simply calling the first matching function. The trouble is the docs mention "search path" which does not apply in our case because both overloads will be in the same schema. And, within a schema, the order of function look up is unclear.

So, I suggest the following actions:
[...]

PR #211, opened 2018-06-07
Make CreateXYZ functions Safe

KevinKelly25:
This pull request is for issue #201 and the CreateXYZ functions were changed to incorporate those changes suggested by @smurthys in that issue thread. 

This is not ready to be approved but I wanted to get opinions on how to handle the changed test script for the new CreateXYZ functions. All I did was change the test script `testClassDBRolesMgmt.sql` so that the tests on the passwords were testing on the initial password. However, this should raise 3 warnings for each test for a total of 9 warnings. It may be useful to see that a warning pops up to see that the new addition is working correctly but I am not sure it is worth doing 9 times. 

Maybe we should change it so each test function incorrectly supplies a password once. Also suppressing the warning might be a good idea since it will be testing the password to make sure it was set to default later.

### DIFF
--- a/src/db/core/addClassDBRolesMgmtCore.sql
+++ b/src/db/core/addClassDBRolesMgmtCore.sql
@@ -133,15 +133,6 @@ CREATE OR REPLACE FUNCTION
    RETURNS VOID AS
 $$
 BEGIN
-   --If password was supplied give warning
-   IF ($7 IS NOT NULL) THEN
-      RAISE WARNING 'For security reasons, parameter "initialPwd" is ignored '
-                    'and user password has been set using the default password '
-                    'policy described in the API documentation.'
-         USING HINT = 'Parameter "initialPwd" will be dropped in the next API'
-                      'release. Please update your code.';
-   END IF;
-
    --record ClassDB role
    PERFORM ClassDB.createRole($1, $2, FALSE, $3, $4, $5, $6);
 
@@ -161,6 +152,14 @@ BEGIN
                   ' ClassDB_Instructor', $3);
    EXECUTE FORMAT('ALTER DEFAULT PRIVILEGES FOR ROLE %s IN SCHEMA %s'
                   ' GRANT SELECT ON TABLES TO ClassDB_Instructor', $1, $3);
+
+   --If password was supplied give warning
+   IF ($7 IS NOT NULL) THEN
+   RAISE WARNING 'parameter "initialPwd" ignored and password set to default value'
+         USING DETAIL = 'Parameter "initialPwd" is deprecated and will be '
+         'dropped in the next release. Please update your code.',
+         HINT = 'Consult the API documentation for details on the password policy.';
+   END IF;
 END;
 $$ LANGUAGE plpgsql
    SECURITY DEFINER;
@@ -313,15 +312,6 @@ CREATE OR REPLACE FUNCTION
    RETURNS VOID AS
 $$
 BEGIN
-   --If password was supplied give warning
-   IF ($7 IS NOT NULL) THEN
-      RAISE WARNING 'For security reasons, parameter "initialPwd" is ignored '
-                    'and user password has been set using the default password '
-                    'policy described in the API documentation.'
-         USING HINT = 'Parameter "initialPwd" will be dropped in the next API'
-                      'release. Please update your code.';
-   END IF;
-
    --record ClassDB role
    PERFORM ClassDB.createRole($1, $2, FALSE, $3, $4, $5, $6);
 
@@ -331,6 +321,14 @@ BEGIN
    --set privileges on future tables the instructor creates in 'public' schema
    EXECUTE format('ALTER DEFAULT PRIVILEGES FOR ROLE %s IN SCHEMA public GRANT'
                || ' SELECT ON TABLES TO PUBLIC', $1);
+
+   --If password was supplied give warning
+   IF ($7 IS NOT NULL) THEN
+   RAISE WARNING 'parameter "initialPwd" ignored and password set to default value'
+         USING DETAIL = 'Parameter "initialPwd" is deprecated and will be '
+         'dropped in the next release. Please update your code.',
+         HINT = 'Consult the API documentation for details on the password policy.';
+   END IF;
 END;
 $$ LANGUAGE plpgsql
    SECURITY DEFINER;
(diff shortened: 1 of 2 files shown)


smurthys:
Thanks @KevinKelly25 for the commits.

I recommend leaving out the last parameter in calls to function `createRole` (instead of explicitly passing parameter 1). This approach permits us to easily change function `createRole`.

Also, please use the following code to raise warning:
```sql
RAISE WARNING 'For security reasons, parameter "initialPwd" is ignored and user password has been set using the default password policy described in the API documentation.' 
      USING HINT = 'Parameter "initialPwd" will be dropped in the next API release. Please update your code.';
```

KevinKelly25:
@smurthys Thank you for the suggested additions. My latest commit reflects those suggestions.

srrollo:
The functions seem to work well in casual usage, however the unit tests fail in several places. There's a bit too much log to dump into this comment, I will try and narrow down where the tests are failed. All the errors are error code 3.

afig:
The error creating `testStu4` is actually a bug with the tests. I will log an issue for this, but essentially, the tests do not perform necessary cleanup if they exit with a `FAIL`. (Eventually they will cleanup because of the `ROLLBACK` at the end of the script, but as is occurring here, they can inadvertently cause other tests to fail.)

Error code 3 occurs when users are not created with the expected password. I will investigate this later if needed.

KevinKelly25:
I haven't tested it yet but I suspect these errors are due to the fact that I just copied and pasted the username into the fields where there was a custom password that needed to be changed. However, I think this leads to an error because of the capitalization and the fail in the 'pg_temp.checkEncryptedPwd()` . I believe that I just need to make sure they are all lowercase.

smurthys:
The test scripts needs to have two sections:

In section 1 ("nominal section"):
- Revise existing creation calls to leave out the password parameter to test the API that way it should be used
- Check passwords against `ClassDB.foldPgID(username)`, because the default pwd policy implemented in function `createRole` sets `ClassDB.foldPgID(username)` as the pwd

In section 2:
-  Create a student, instructor, and dbman with a custom pwd and expect to see warnings only for these three creation attempts. Nothing in section 1 should cause a warning
- Check password as in section 1 to verify that the new API ignores custom pwd
- This section should be clearly marked and set off from section 1 so it can be removed easily in the next release (when the pwd parameter will be removed)

smurthys:
After reading the [error-message guide](https://www.postgresql.org/docs/9.6/static/error-style-guide.html) @afig helpfully provided in [another thread](https://github.com/DASSL/ClassDB/pull/203#issuecomment-395596758), I feel the following code is better to raise warning:

```sql
RAISE WARNING 'parameter "initialPwd" ignored and password set to default value'
      USING DETAIL = 'Parameter "initialPwd" is deprecated and will be dropped in the next release. Please update your code.',
      HINT = 'Consult the API documentation for details on the password policy.';
```

KevinKelly25:
With the new commits the new format for `testClassDBRolesMgmt.sql` follows what @smurthys suggested above. The warning has also been moved to to be at the end of the functions. 

The code has been tested and results match what is expected. IE, Section 1 of the tests runs smoothly and without warnings. Section 2 tests each CreateXYZ function with a non-Null password and then checks that each test role was actually set to the ClassDB.FoldPgID(username). This also creates 3 warnings which is intended for each function.

Also, Section 2 is clearly marked at the end of the test script so that it can be easily removed when no longer needed.

This PR is ready for reviews.

smurthys:
Thanks for the changes @KevinKelly25. They look good.

If it is not too much effort, would you please change the hint text as follows?
`Consult ClassDB documentation for details on password policy.`

I like the new section in `testClassDBRolesMgmt.sql` and have the following observations:
- The section can benefit from a comment that it should be removed when parameter `intiailPwd` is removed from functions `createXYZ`
- L1001 has an extra leading space
- L1018-L1029: Remove this code because the later ROLLBACK does that automatically. Such code elsewhere in this script should also be removed, but that is not the focus of this PR. Instead, I suggest focusing on getting just the new code right.
- L1035 should be `RAISE INFO` instead of `SELECT`. L976 provides a model.
- Consider renaming the function `defaultPasswordTest` to something like `rejectCustomPasswordTest` to better communicate intent.
- I recommend using more unique role names in calls to createXYZ, for example, `testStuCustomPwd`

afig:
This PR is looking pretty good, thanks @KevinKelly25.

In addition to the comments made by @smurthys , I would only add that we may want to slightly change the description of section 2, since the tests are not actually verifying that a `WARNING` is raised, only that the passwords are rejected. Currently it reads:

>tests each of the createXYZ functions to make sure that if supplied with a password that the *function raises a warning and* sets the password to the default

KevinKelly25:
Thank you @smurthys  and @afig for the suggestions. I addressed those issues in the last commits.

KevinKelly25:
Thanks @afig, fixed that small issue

KevinKelly25:
@smurthys thank you for the comments. They will be included in the next commit. 

I know that section 2 is just temporary until a later update to the createXYZ functions. However, after running the test script many times I wonder if it may be better to RAISE a notice at the beginning of the `pg_temp.rejectCustomPasswordTest()`? 

This way the user knows that the 3 warnings are normal and there is less chance of a user misunderstanding the test and thinking that is may have failed in some way.

Adding something along the lines of:
`RAISE NOTICE 'The following test should RAISE three warnings';`

smurthys:
I like the idea of raising a notice as @KevinKelly25 suggests. I wonder if others sees a downside.

smurthys:
Just curious if @afig or @srrollo have any thoughts on raising a notice as @KevinKelly25 suggests.

afig:
I think raising a `NOTICE` is a good idea. The sample message provided looks pretty good, I would just add something that states the `NOTICE`s should be related to password rejection.

So adapting @KevinKelly25's examples, something like:

```sql
RAISE NOTICE 'The following test should RAISE three warnings regarding ignoring of an initial password';
```

KevinKelly25:
Thank you for the feedback. I added the notice to the last commit.

srrollo:
I think the changes look good. Just want to re-afirm my approval.

afig:
Looks good. I can't test right now, but not sure if a space is needed at the beginning of the second line of the warning. I think the concatenation does not insert a space, so when issued, the warning will print out `regardingignoring` as one word.

KevinKelly25:
@afig thank you for the catch! I didn't notice that when testing.

PR #216, opened 2018-06-10
Define server team role

smurthys:
The commits in this PR add a new server role `ClassDB_Team` in preparation for team support. User and ClassDB role management need the server role to support teams. (Those changes are and should be subject of another PR).


smurthys:
Thank you all for reviewing. I have addressed the omission @KevinKelly25 pointed out.

Issue #220, opened 2018-06-12
No functionality available to drop a team (N)


afig:
From an earlier version of M3's Wiki page:

- [ ] Add function `dropTeam` to `addClassDBRolesMgmtCore.sql`
  - Same signature as `dropStudent`, except first parameter is `teamName`
  - Same functionality as `dropStudent` except call `revokeTeam` instead of `revokeStudent`
- [ ] Add function `dropAllTeams` to `addClassDBRolesMgmtCore.sql`

Issue #219, opened 2018-06-12
No functionality available to revoke a team (N)


afig:
From an earlier version of M3's Wiki page:

- [ ] Revise function `revokeClassDBRole` to test isRoleKnown() instead of testing isUser()
- [ ] Add function `revokeTeam` to `addClassDBRolesMgmtCore.sql`
  - Similar functionality as `revokeStudent` except no changing connection limit or statement timeout
  - EDIT: We also have to remove any default privileges that were added (see issue #221).

Issue #218, opened 2018-06-12
No functionality available to create a team (N)


afig:
From an earlier version of M3's Wiki page:

- [ ] Add function `createTeam` to `addClassDBRolesMgmtCore.sql`
  - Same signature as `createStudent`, except first parameter is `teamName` and `fullName` is optional
  - Call `createRole` to create server role: ~~no change anticipated to `createRole`~~ Changes required: see comments below.
  - Grant role `ClassDB_Team` to the team role
  - Grant instructors privileges to the team's schema: exactly (?) as is done with new student roles
  - No need to set connection limit or statement timeout

afig:
Functionality has been implemented along with basic tests in branch `add-teams-functionality`, however a change to `createRole()` may be needed (see Issue #223)

afig:
Having the `fullName` parameter for the future `createTeam()` be optional results in a conflict with the current implementation of `createRole()`. This is because `createRole()` raises an `EXCEPTION` if `fullName` is NULL or empty.

A few possible solutions are:

1. Have `createRole()` not enforce a non-empty `fullName` if `isTeam` is true. This will require it to be ensured that allowing a `NULL` or empty name does not lead to other issues. The [ClassDB Role-Creation Process Diagram](https://github.com/DASSL/ClassDB/wiki/Adding-Users#additional-info-and-troubleshooting) will also have to be updated.

2. Make `fullName` for teams required. Note that it may be slightly confusing for users to be required to enter two names for a team.

3. Leave `fullName` "optional", but coalesce with `teamName` before calling `createRole()`. 

The first solution is ideal, and is likely not be a huge effort to implement (there would not be any API changes since the behavior only impacts teams, which were not officially supported in 2.0). Solutions 2 and 3 are left as "backups" in the event of time constraints or issues with the first solution. Current development of teams is using solution 3 as a temporary measure so as to not slow down development.

smurthys:
Option 3 in @afig's list works, but should probably be implemented within function `createRole` so that function can be evolved later. 

To be specific, `createRole` can use a team's role name as full name if the full name supplied is empty (after trimming) or `NULL`.

afig:
I like @smurthys's suggestion about implementing option 3, but in `createRole()`. This opens up a couple of questions, however:

[...]

PR #226, opened 2018-06-14
Add team management

afig:
This PR focuses on adding functions to create, revoke, and drop teams in ClassDB. In particular, it leaves out functionality that manages team membership, which will be added in a separate PR due to the large number of changes.

Unit tests for all 4 functions have been created, and all pass on my instance except for `testCreateTeam()`, which fails with Code 4 due to a pending decision related to optional `fullName`s (see Issue #218). Currently, if a `fullName` is not provided, it is set to `teamName` (final behavior will depend on the decision made in #218, as such, **do not approve this PR** until then).

As @smurthys discussed in PR #211, the tests in `testClassDBRolesMgmt.sql` need several improvements in relation to avoiding unnecessary operations and improving consistency. The new tests were created with these improvements in mind, but the rest of the test script has been left unchanged, leaving the improvements for future development.

Changes to privilege tests will wait for team membership functionality to be added.

Closes: #218, Closes: #219, Closes: #220

### DIFF
--- a/src/db/core/addClassDBRolesMgmtCore.sql
+++ b/src/db/core/addClassDBRolesMgmtCore.sql
@@ -520,4 +520,57 @@ GRANT EXECUTE ON FUNCTION
 
 
 
+--Define function to register a team and perform corresponding configuration
+--Calls ClassDB.createRole with corresponding parameters
+--Grants appropriate privileges to newly established role and schema
+CREATE OR REPLACE FUNCTION
+   ClassDB.createTeam(teamName ClassDB.IDNameDomain,
+                      fullName ClassDB.RoleBase.FullName%Type DEFAULT NULL,
+                      schemaName ClassDB.IDNameDomain DEFAULT NULL,
+                      extraInfo ClassDB.RoleBase.ExtraInfo%Type DEFAULT NULL,
+                      okIfRoleExists BOOLEAN DEFAULT TRUE,
+                      okIfSchemaExists BOOLEAN DEFAULT TRUE)
+   RETURNS VOID AS
+$$
+BEGIN
+   --record ClassDB role
+   PERFORM ClassDB.createRole($1, COALESCE($2, teamName), TRUE, $3, $4, $5, $6);
+   
+   --get name of role's schema (possibly not the original value of schemaName)
+   $3 = ClassDB.getSchemaName($1);
+   
+   --grant server-level team group role to new team
+   PERFORM ClassDB.grantRole('ClassDB_Team', $1);
+   
+   --grant instructors privileges to the team's schema
+   EXECUTE FORMAT('GRANT USAGE ON SCHEMA %s TO ClassDB_Instructor', $3);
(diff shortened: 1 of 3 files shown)


smurthys:
Thanks for making these enhancements @afig. It looks like only a small amount of work is left to fully support teams.

afig:
I just pushed a commit that addresses the issue of not allow full names to be optional. This involved slightly modifying the behavior of `createRole()` and a `CHECK` constraint in the `RoleBase` table (see discussion at issue #218) for more information.

The order that the arguments are verified in `createRole()` had to be slightly modified so that `isTeam` is validated before using it to check whether or not to validate `fullName`

I think this PR is ready for a final review. As mentioned earlier, functionality that allows team members to be added and removed will be added in a separate PR.

srrollo:
Starting reviewing this and it looks good so far. I ran into an issue with the tests crashing because `NOT NULL` was still being universally enforced for 'FullName' in `ClassDB.RoleBase`. I believe this issue is that if you apply the changes to an existing system, `RoleBase` will (correctly) not be replaced.

This is similar to the upgrade issues with connection logging, and I propose the following fix. Add:
```sql
   ALTER TABLE ClassDB.RoleBase ALTER COLUMN FullName DROP NOT NULL;
   ALTER TABLE ClassDB.RoleBase ADD CONSTRAINT EnforceFullName CHECK(isTeam OR (TRIM(FullName) <> '' AND FullName IS NOT NULL));
```
before `ClassDB.RoleBase` is created with a comment stating this is for 2.0 to 2.x upgrade, etc. (This appears to work on my system). I think this will need some logic to check if `RoleBase` exists and if the new constraint exists. `DROP NOT NULL` is idempotent, but ADD CONSTRAINT fails if the constraint already exists.

afig:
Thanks for pointing that out @srrollo, I had not tried running the scripts on a DB that already had ClassDB installed. I also noticed another change that needed an upgrade statement, so I added that.

Thankfully, both `ALTER TABLE` and `... DROP CONSTRAINT` have `IF EXISTS` variants, so nothing procedural needed to be written. `rolebase_fullname_check` is the default name for the constraint, so to keep it simple, I kept that name even when adding the constraint separately.

smurthys:
Excellent enhancement @afig. I have eyeballed the code and don't see any issue. I will run the scripts in my installation in a few hours and report back.

srrollo:
Nice find on the `IF EXISTS` clauses, it's quite nice to have all of those. I think this looks good to go.

smurthys:
The changed scripts compile and the tests run successfully. 

I find only two minor issues in `testClassDBRolesMgmt.sql`. Other than these, this PR seems is ready to merge.
- L743: Incorrect comment, likely a copy/paste error
- L1130 and L1140: Seems to have an extra tab space before `NOT`

afig:
Thanks for pointing out those comment issues @smurthys, they have been addressed in the most recent commit.

PR #232, opened 2018-06-17
Add functions to get and compare server versions

### LINKED ISSUE #231: Helper functions needed to compare server versions (N)
Many tasks related to compatibility with Postgres version prior to 9.6 are based on server version, for example the following issues: #228, #229, #230

smurthys:
This PR adds the following functions:
- `getServerSetting` get any server setting by name
- `getServerVersion` to get server's version number
- `compareServerVersion` to compare a version number with current server's, or compare any two version numbers

I am still working on a test script for these changes.

Closes #231

### DIFF
--- a/src/db/core/addHelpersCore.sql
+++ b/src/db/core/addHelpersCore.sql
@@ -555,6 +555,8 @@ $$ LANGUAGE plpgsql
 ALTER FUNCTION
    ClassDB.compareServerVersion(VARCHAR, VARCHAR, BOOLEAN) OWNER TO ClassDB;
 
+--Limit to ClassDB to prevent exceptions due to incorrect args
+-- too much development effort to prevent all possible exceptions
 REVOKE ALL ON FUNCTION
    ClassDB.compareServerVersion(VARCHAR, VARCHAR, BOOLEAN) FROM PUBLIC;
 
--- a/src/db/core/addHelpersCore.sql
+++ b/src/db/core/addHelpersCore.sql
@@ -461,29 +461,119 @@ ALTER FUNCTION
 
 
 
---Define a function to retrieve specific capabilities a user has
--- use this function to get status of different capabilities in one call
+--Define a function to get the value of any server setting
+--Queries the catalog view pg_settings to avoid exceptions if the setting name
+-- supplied is not found
+--Returns NULL if the setting name supplied is not found
+CREATE OR REPLACE FUNCTION
+   ClassDB.getServerSetting(settingName VARCHAR)
+   RETURNS VARCHAR AS
+$$
+   SELECT setting FROM pg_catalog.pg_settings
+   WHERE name = $1;
+$$ LANGUAGE sql
+   RETURNS NULL ON NULL INPUT;
 
---Commenting out the function because a unit test is yet to be developed
---CREATE OR REPLACE FUNCTION
---   ClassDB.getRoleCapabilities(roleName ClassDB.IDNameDomain,
---                               OUT isSuperUser BOOLEAN,
---                               OUT hasCreateRole BOOLEAN,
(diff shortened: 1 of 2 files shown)


smurthys:
I just pushed changes that test the new helper functions. I welcome your review.

srrollo:
All of the tests pass on my system, and the function work well overall. I found one issue that I think is best addressed in `getServerVersion`. On my Postgres distribution from Ubuntu, `getServerVersion()` returns `'10.3 (Ubuntu 10.3-1)'`. This causes an error when using `compareServerVersion()`: `ERROR:  invalid input syntax for integer: "3 (Ubuntu 10"`. It seems the best solution would be to have `getServerVersion()` throw out anything after it extracts the server version. I don't think we need to have `compareServerVersion()` handle this specifically, because we can document that it expects a normally formatted version number as input.

smurthys:
Thanks @srrollo for sharing your Ubuntu results. I am testing with pg 9.6.3 on Windows 10. I wonder if a later version of pg formats the result differently in that it includes OS information.

@KevinKelly25 are you able to test the changes in this PR in your environment? I assume you have pg9.3 and 10.4 on Windows.

I agree with @srrollo's suggestion that `getServerVersion` should remove any extra info. I'll await @KevinKelly25's response before designing a solution.

srrollo:
I just tested my BigSQL 9.6.9 installation, and `getServerVersion()` returns `9.6.9`. I suspect `(Ubuntu 10.3-1)` is "compiled in" to the Ubuntu version, and is not strictly related to the OS. This is pretty common on software provided in Linux distro repos, in order to show that it is software version X compiled by distro Y for package version Z.

KevinKelly25:
For 10.4 and 9.3 the tests all pass with no issues.

afig:
I tested this with Postgres 9.6 on Windows and on a Ubuntu instance and they worked correctly. Unlike @srrollo's Ubuntu installation, the instance I have installed came from Postgres' own distribution, rather than Ubuntu's. This would agree with description that @srrollo gave about packaging.

smurthys:
Thank you all for reporting the results from your installations.

Can we safely say that a custom distro returns a value of the form `x.y.z (os info)` in that we can safely stop if/when we see an open parenthesis?

smurthys:
I just pushed changes to both the helper functions and the tests to account for distro suffix in version number. With these changes, both `getServerVersion` and `compareServerVersion` consider only the portion before opening parenthesis in a retrieved/supplied version number.

KevinKelly25:
Tests for version 9.3 and 10.4 still working. @smurthys thanks for the useful helper functions and tests

smurthys:
I see the test script has an issue: L195 in `testHelpers.sql` needs some attention. I will look at it in an hour or so.

smurthys:
The change I just pushed should address the issue on L195 in `testHelpers.sql`. Before this change that line would have failed in 3rd party distros.

smurthys:
I just pushed a few changes which collectively do the following:
- Change the default value of parameter `testPart2` to `TRUE` from `FALSE` in `compareServer` functions. I made this change because many of our choices depend on part 2 of version number.
- Add three shortcut functions `isServerVersionBefore`, `isServerVersionAfter`, and `isServerVersion` to test if the current server's version precedes, succeeds, or matches another version number.
- Change unit tests to match the change to the default value of parameter `testPart2`
- Add unit tests for shortcut functions

I believe at this point the changes in this PR pretty much define the PR's scope.

smurthys:
I agree we need to check the distros @srrollo lists. 

We should probably also list the distros we have tested on, or we should specifically say we have tested only on 1st party Linux and Windows distros.

smurthys:
Thank you all for reviewing.

Issue #227, opened 2018-06-16
Make ClassDB compatibile with Postgres versions prior to 9.6


smurthys:
All scripts need to be tested in different versions of Postgres from 9.3 to 9.5 to verify compatibility status with Postgres versions prior to 9.6. Ideally, ClassDB should be made compatible with Postgres 9.3. Where compatibility is not possible, workarounds should be explored and documented.

srrollo:
Not sure how much progress has been made on these issues, however it looks like Postgres 9.3 goes EOL in September ([support timeline](https://www.postgresql.org/support/versioning/)). Considering this, it may not be worth adding 9.3 specific workarounds (depending on how much effort it will be).

smurthys:
Good point @srrollo about EOL of pg 9.3. However, I think it makes sense to support it because we should expect those still using it to use it for another year or so. 

Of course, the choice also depends on how much effort it is to make ClassDB compatible with 9.3. One way to think about this might be that 9.3 compatibility is lowest priority among all compatibility issues.

KevinKelly25:
As seen in #233 there is a chance that this epic is starting to grow beyond the scope and time frame of M3.

We already have 5 outstanding issues in this epic and although we have made progress on those issues the amount of dev effort required for the full compatibility of pre-9.6 versions keeps rising. Also, there are large parts of ClassDB we have not been able to implement on 9.3 because of the outstanding issues. I have a suspicion that after we fix some or all of these issues that other issues are going to pop up in those portions of code that we have not been able to get to yet. 

With that in mind and that we are trying to get M3 done relatively soon, I suggest that we table the epic for a later milestone so we can put more priority into the main goals of M3. I am curious to know other opinions on if we should include this in M3, a later milestone or that it isn't worth doing at all.

smurthys:
Recapping the key lesson from PR #233 that static SQL code cannot use any unsupported feature/syntax even if the code is conditionally executed only on supported versions. This means parts of scripts need to be executed either dynamically or be duplicated to create version-specific code. The former is approach is bearable in some cases; the latter should be avoided in most cases due to maintenance concerns.

With that lesson in the background, we need to **immediately** answer at least the following questions.

1. Should pre9.6 compatibility be in M3? Or, is it better create an M4 for this summer with only pre9.6 compatibility as a goal?
2. Taking a cue from @srrollo's comment earlier, should we put the effort to support pg9.3 given that it reaches EOL in this September (2018)?
[...]

PR #233, opened 2018-06-18
Guard IF NOT EXISTS in CREATE INDEX for pre-pg9.5

### LINKED ISSUE #230: Guard uses of IF NOT EXISTS in CREATE INDEX: changed in pg9.5 (N)
Prior to 9.5 CREATE UNIQUE INDEX does not support IF NOT EXISTS. This causes a problem in `addRoleBaseMgmtCore.sql` LN 61, when trying to create a unique index on folded RoleName.

```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_Unique_FoldedRoleName
ON ClassDB.RoleBase(ClassDB.foldPgID(RoleName));
```
Error:
```psql
psql:core/addRoleBaseMgmtCore.sql:62: ERROR:  syntax error at or near "NOT"
LINE 1: CREATE UNIQUE INDEX IF NOT EXISTS idx_Unique_FoldedRoleName
```

As seen [here](https://www.postgresql.org/docs/9.5/static/sql-createindex.html) in the docs IF NOT EXISTS is an modifier added to the functionality of CREATE UNIQUE INDEX in 9.5

smurthys:
The changes in this PR guard the use of `IF NOT EXISTS` in `CREATE INDEX`. 

The changes work in pg9.6, **but need to be tested in pg9.3 or 9.4 as well as in 10.x**. 

Fixes #230 

BTW, these changes give a glimpse of the code structure we might need to make the product compatible with Postgre versions prior to 9.6

### DIFF
--- a/src/db/core/addHelpersCore.sql
+++ b/src/db/core/addHelpersCore.sql
@@ -461,29 +461,182 @@ ALTER FUNCTION
 
 
 
---Define a function to retrieve specific capabilities a user has
--- use this function to get status of different capabilities in one call
+--Define a function to get the value of any server setting
+--Queries the catalog view pg_settings to avoid exceptions if the setting name
+-- supplied is not found
+--Returns NULL if the setting name supplied is not found
+CREATE OR REPLACE FUNCTION
+   ClassDB.getServerSetting(settingName VARCHAR)
+   RETURNS VARCHAR AS
+$$
+   SELECT setting FROM pg_catalog.pg_settings
+   WHERE name = $1;
+$$ LANGUAGE sql
+   RETURNS NULL ON NULL INPUT;
 
---Commenting out the function because a unit test is yet to be developed
---CREATE OR REPLACE FUNCTION
---   ClassDB.getRoleCapabilities(roleName ClassDB.IDNameDomain,
---                               OUT isSuperUser BOOLEAN,
---                               OUT hasCreateRole BOOLEAN,
---                               OUT canCreateDatabase BOOLEAN)
---   AS
---$$
---BEGIN
---   SELECT rolsuper, rolcreaterole, rolcreatedb FROM pg_catalog.pg_roles
---   WHERE rolname = $1;
---END;
---$$ LANGUAGE plpgsql;
+ALTER FUNCTION ClassDB.getServerSetting(VARCHAR) OWNER TO ClassDB;
 
---ALTER FUNCTION
---   ClassDB.getRoleCapabilities(roleName ClassDB.IDNameDomain,
---                               OUT isSuperUser BOOLEAN,
---                               OUT hasCreateRole BOOLEAN,
---                               OUT canCreateDatabase BOOLEAN)
---   OWNER TO ClassDB;
+
+
+--Define a function to get the server's version number
+--Removes additional info a distro may have suffixed to the version number
+-- e.g., Ubuntu's distro is known to return '10.3 (Ubuntu 10.3-1)', whereas
+-- a Postgres distro returns just '10.3'
+CREATE OR REPLACE FUNCTION ClassDB.getServerVersion()
+   RETURNS VARCHAR AS
+$$
+   --get value of setting 'server_version' and remove any distro-added suffix
+   SELECT TRIM(split_part(ClassDB.getServerSetting('server_version'), '(', 1));
+$$ LANGUAGE sql
+   RETURNS NULL ON NULL INPUT;
+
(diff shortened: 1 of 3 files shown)


srrollo:
I haven't had to closely examine the code yet, however the version of ClassDB this branch provides installs, and the tests pass, on 10.3.

KevinKelly25:
10.4 works but 9.3 is still getting
```psql
psql:core/addRoleBaseMgmtCore.sql:85: ERROR:  syntax error at or near "NOT"
LINE 18:       CREATE UNIQUE INDEX IF NOT EXISTS idx_Unique_FoldedRol...
````

I was doing some testing and I think it might be that even though 9.3 never could get there it still does a syntax check on the whole code block before running through it. I tested this out by making a small block of code that was impossible to get to and contained syntax that 9.3 doesn't use. Even though it was not possible to get to that code it still would get an error every time I ran that code. I am still testing to see if there is a way to get around this.

**Comment by afig** [comment]
Nice catch and analysis @KevinKelly25. Looks like there is a workaround. We can tell Postgres that the "invalid" statement is a [dynamic command](https://www.postgresql.org/docs/9.3/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-EXECUTING-DYN), which means that it will not attempt to cache it, avoiding the syntax error.

L79-L83 of `addRoleBaseMgmtCore.sql` would look like:

```sql
ELSE
   --works on pg9.5 or later
   EXECUTE 'CREATE UNIQUE INDEX IF NOT EXISTS idx_Unique_FoldedRoleName'
            ' ON ClassDB.RoleBase(ClassDB.foldPgID(RoleName))';
END IF;
```

smurthys:
Thanks @KevinKelly25 for the report. I was quite expecting this outcome. 😢 

The dynamic-sql alternative @afig suggests is an alternative, but we should in general prefer static code over dynamic code, because dynamic code is prone to errors and hard to debug. They do have the advantage that the alternative "code" can stay inline and then later be easily replaced with static code.

Another alternative may be to place any code that works only in a later version inside a temp function, because I believe a function body is late bound. Just so we can verify, **please run the following code on pg9.3**. Whether we use that approach or not, I am curious to know the outcome.

However, this issue brings in to focus our priority for M3, and also calls for us to make a a long-term strategy on supporting pg versions prior to 9.5. I propose we have that discussion in #227 so our thinking is recorded in one place and remains public.

```sql
CREATE OR REPLACE FUNCTION pg_temp.createIndexUniqueFoldedRoleName_afterPg94() 
RETURNS VOID AS
$$
   CREATE UNIQUE INDEX IF NOT EXISTS idx_Unique_FoldedRoleName
   ON ClassDB.RoleBase(ClassDB.foldPgID(RoleName));
$$ LANGUAGE sql;

DO
$$
BEGIN
   IF ClassDB.isServerVersionAfter('9.4') THEN
      PERFORM pg_temp.createIndexUniqueFoldedRoleName_afterPg94();
   ELSE
      --pg9.4
      --works on any pg version, but the post pg9.4 version is much simpler
      IF NOT EXISTS (SELECT indexname FROM pg_catalog.pg_indexes
                     WHERE schemaname = 'classdb'
                          AND tablename = 'rolebase'
                          AND indexname = 'idx_unique_foldedrolename'
                    )
      THEN
         CREATE UNIQUE INDEX idx_Unique_FoldedRoleName
         ON ClassDB.RoleBase(ClassDB.foldPgID(RoleName));
      END IF;
   END IF;
END
$$;
```

KevinKelly25:
I ran the previous temp function @smurthys brought up on 9.3. It is still giving the same error upon trying to create the function.

smurthys:
Thanks @KevinKelly25 for the confirmation. 

This confirmation leaves us with two options as far as this issue is concerned:
- Implement just the pre9.5 solution because it works in all version
- Implement different versions for pre9.5 and 9.5+ with the 9.5+ version being a dynamic query

Either approach is acceptable. However, my concern is this issue is not representative of the requirements/complexity/effort needed to support compatibility with previous pg versions. I invite a general discussion at #227.

smurthys:
Given the decision in #227 to address Postgres compatibility in M4, should we abandon this PR? 

It will be clean to address all compatibility issues in M4, and there isn't much effort lost if we abandon now. (I will just stash the work for later use.)

srrollo:
I think it makes sense to abandon this PR until M4, since we need to re-implement the changes using one of the options outlined by @smurthys. I agree we don't lose much in terms of code already produced.

afig:
Closing this PR without merging makes sense. We can reevaluate the solution to compatability problems later, during M4.

PR #239, opened 2018-06-21
Add Team Views

### LINKED ISSUE #236: Define team-related views (N)
New view definitions are needed to support teams feature. They should probably be added to `addClassDBRolesViewsCore.sql`.

- [ ] `Team`: list teams (similar to view `User` but with only basic team information)
  - Fields `TeamName`, `FullName`, `SchemaName`, `ExtraInfo`, `MemberCount`
- [ ] `TeamMember`: list teams and members
  - Fields TeamName, MemberName

These views probably need more fields, but the needs are yet unclear.

KevinKelly25:
Adds the team views: ClassDB.Team and ClassDB.TeamMember. Fixes #236 

The view TeamMember displays all roles that are added to a team and their respective team(s).

The view Team displays all teams with their attributes and an aggregate `MemberCount`.

In order to test this properly the `ClassDB.addToTeam` function from branch `add-teams-membership` will need to be added.

I added it manually to my local ClassDB and everything seems to working properly.

### DIFF
--- a/src/db/core/addClassDBRolesViewsCore.sql
+++ b/src/db/core/addClassDBRolesViewsCore.sql
@@ -135,7 +135,7 @@ GRANT SELECT ON ClassDB.DBManager TO ClassDB_Instructor, ClassDB_DBManager;
 -- creates a derived table containing all teams for a self-join
 CREATE OR REPLACE VIEW ClassDB.TeamMember AS
   SELECT Team.RoleName Team, Member.RoleName Member
-  FROM Classdb.RoleBase Member
+  FROM ClassDB.RoleBase Member
   JOIN ClassDB.RoleBase Team ON ClassDB.isMember(Member.RoleName, Team.RoleName)
   WHERE NOT Member.IsTeam AND Team.IsTeam
   ORDER BY Team;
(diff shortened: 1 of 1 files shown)


srrollo:
The code looks good overall. `ClassDB.Team` appears to not display teams that have no members, however. For example:
```sql
SELECT ClassDB.createTeam('MyTeam');
SELECT * FROM ClassDB.Team;
```
yields:
```sql
 teamname | fullname | schemaname | extrainfo | membercount
----------+----------+------------+-----------+-------------
(0 rows)
```

At a glance, I think the second change suggested by @smurthys would fix this by removing the inner join.

KevinKelly25:
@smurthys Thanks for the optimization idea, I used that method for both of the views in the new commit. 

@srrollo Thanks for spotting that. As you said, with the new view the error does not persist.

smurthys:
These changes compile and run correctly. 

A tiny correction is possible on L138: change `Classdb` to `ClassDB`. Other than that, this PR seems ready to merge.

KevinKelly25:
Thanks @smurthys for finding the capitalization error. 

I fixed that in the latest commit and also changed a comment to better fit the updated view structure.

KevinKelly25:
Thank you all for the reviews. I fixed the spacing in last commit @afig. Thanks for pointing that out.

PR #240, opened 2018-06-22
Add ddl session logging

### LINKED ISSUE #237: Add session id to table DDLActivity (N)
Adding column `SessionID` to table `DDLActivity` makes it possible to connect each DDL acitivity to an application and a connection. This information is necessary for analysis operations.

Postgres doesn't have a function to get session id, but the following query (based on [pg doc for `log_line_prefix`](https://www.postgresql.org/docs/9.6/static/runtime-config-logging.html)) should work:

```sql
SELECT to_hex(trunc(EXTRACT(EPOCH FROM backend_start))::integer) || '.' ||
       to_hex(pid)
FROM pg_stat_activity
WHERE pid = pg_backend_pid();
```

### LINKED ISSUE #243: SET SESSION AUTHORIZATION breaks ClassDB.getSessionID() (W)
Observed in `testClassDBRolesMgmt.sql`. This is because only superusers can view `backend_start` in `pg_stat_activity` for other users. This is an issue because DDL operations performed while `SET SESSION AUTHORIZATION` is active fail because the DDL activity trigger tries to insert a `NULL SessionID` into `DDLActivity`. Below is an example error message from trying to perform a DDL op. under `SET SESSION AUTHORIZATION`:
```sql
SET SESSION AUTHORIZATION sysadm;
classdb200=> CREATE TABLE ttable(a INT);
ERROR:  null value in column "sessionid" violates not-null constraint
DETAIL:  Failing row contains (sysadm, 2018-06-22 15:29:47.738855, CREATE TABLE, sysadm.ttable, null).
CONTEXT:  SQL statement "INSERT INTO ClassDB.DDLActivity VALUES
         (SESSION_USER, statement_timestamp() AT TIME ZONE 'utc', TG_TAG, objId,
            ClassDB.getSessionID())"
PL/pgSQL function logddlactivity() line 34 at SQL statement
```

There are three solutions I see for this issue:
1. Change `ClassDB.getSessionID()` owner to superuser, and set it to `SECURITY DEFINER`. This will allow it to always return a session ID. However the `SessionID` will be "wrong" under testing conditions in the sense that the session ID stored will be that of the superuser running the test, but the username will be the user specified in `SET SESSION AUTHORIZATION`. This is probably OK since it will only show up in testing. In normal conditions, this will still return the correct session ID.

1. Check if the caller of `ClassDB.SessionID()` is superuser, and if not return the dummy value (`0000000.0000000`) we are using in `ClassDB.ConnectionActivity`. In normal usage, the caller of this function will always be a superuser, since the DDL trigger function is executed as a superuser.

1.  `COALESCE` `backend_start` with `0000000`,

srrollo:
This commit adds logging of `SessionID` in the DDL activity trigger and updates the frequent user views for compatibility. Additionally, a helper function `ClassDB.getSessionID()` was added. Note that this branch contains all commits from #235 due to the nature of the changes. As such, this will be blocked by #235.

Closes #237, Closes #243


smurthys:
Thanks @srrollo for initiating these changes. They look quite good.

Please consider the following observations.

**`addHelpersCore.sql`:**
- Please add a link and comments about the origin of the code in function `getSessionID`. The link and suitable comment text is available in the description of Issue #237 

**`addUserMgmtCore.sql`:**
- As with the discussion on table ConnectionActivity  in PR #235, it is better to re-create table `DDLActivity` if it exists and is empty.

**`addFrequentViewsReco.sql`:**
- In the table returned by function `getUserDDLActivity`, column `SessionID` should be immediately after `StatementStartedAt` for compatibility with function `getUserActivity`. (I wonder if we should also return `ApplicationName`, but then we need to think about join efficiency: so, not now.)

srrollo:
Once we decide on a solution to the upgrade issues pointed out in #235, I will update this branch to use the same solution.

srrollo:
I pushed a candidate ER Schema to this branch that reflects the changes to `ClassDB.ConnectionActivity` and `ClassDB.DDLActivity`. I am wondering, however, if `ClassDB.ConnectionActivity` is no longer a weak entity due to the new PK (or index).

smurthys:
I am glad @srrollo spotted and logged issue #243. The fix also seems adequate. I think a brief end-of-line comment in code on L680 in `addHelpers.sql` is helpful so future maintainers know why this function must be owned by a superuser.

The changes to table `DDLActivity` in `addUserMgmtCore.sql` also look good. I do observe the following  things:
- L59, L89, L97: Column `SessionID` is missing `CHECK` constraint to prevent empty string
- L112: I see one extra indentation space

srrollo:
I have added the `CHECK` constraint to the `SessionID` columns.

srrollo:
I just pulled in the new changes from dev, so the diff should be clean now.

Issue #222, opened 2018-06-12
Add Teams to ClassDB


afig:
Teams allow students to collaborate using shared spaces and are the highlight feature of ClassDB v2.1. The `ClassDB_Team` group role has been defined in PR #216. We now need to implement functions and views to help manage teams.

Team management:
- Creating teams
- Revoking teams
- Dropping teams
- Managing team memberships

Issue #207, opened 2018-06-04
Testing scripts require a user to correctly type in a password (N)


afig:
Several scripts, including ```testAddConnectionActivityLogging.psql``` and all Privilege tests, require the user who is running the tests to enter a password when connecting to the database as the test users/roles. In ```testAddConnectionActivityLogging.psql```, the password for the test roles are ```' '```, and for the privilege tests, the password is the same as the name of the test role.

Users are essentially only given one chance to type in the password correctly. If the user running the tests incorrectly types the password for the test role, the psql client is left in a state of not being connected to a database, requiring the user to quit/exit the psql window and having to reconnect to the database and run the tests anew.

Two possible solutions are:

1. Gracefully handle an incorrectly typed password, allowing the user another chance to type in the password again (security is not an issue here since these are only test scripts with test roles).
2. Have it so users do not have to manually type in a password.

The first solution may not be easy to implement, since this behavior is by design (running a script places psql in non-interactive mode): 

> If the connection attempt failed (wrong user name, access denied, etc.), the previous connection will only be kept if psql is in interactive mode. When executing a non-interactive script, processing will immediately stop with an error. This distinction was chosen as a user convenience against typos on the one hand, and a safety mechanism that scripts are not accidentally acting on the wrong database on the other hand.
_[psql - \connect](https://www.postgresql.org/docs/10/static/app-psql.html)_

The second solution would be ideal, making running of tests significantly easier on the user, and possibly, easier to automate in the future. This should be possible through the use of the [\connect psql command](https://www.postgresql.org/docs/current/static/app-psql.html) with a [Connection String](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING). However, it remains to be investigated how the correct ```host``` and ```port``` settings will be determined, and whether the connection string can be generated at script runtime.

srrollo:
See [here](https://github.com/DASSL/ClassDB/pull/214) for one auto-login solution. psql allows automatic login by setting the `PGPASSWORD` environment variable (in the client system's shell) to the password of the logging in user. psql can also set shell environment variables using the `\setenv` meta command. So by setting `PGPASSWORD` to the necessary password before using `\c`, the user will auto login. At the end, `PGPASSWORD` should be set to nothing, otherwise psql will continue to try and use its value for any user logging in and will not prompt the user for a password.

PR #247, opened 2018-06-23
Update and improve privilege tests (M3)

afig:
This PR updates the privilege tests to also test privileges on newly added functions. The tests are still not perfect, but they are gradually improving.

In addition to expanding the objects tested, the main script has also been updated to not require the user to enter a password for every connection to the database. The tests also store who the original user was rather than assuming it was the `postgres` role. Thanks to work that @srrollo and @smurthys performed in PR #214 for these fixes.

Closes: #207, Closes: #222


afig:
Thanks for the reviews everyone.

PR #253, opened 2018-06-23
Added m3 docs

srrollo:
Updates the repo docs for m3


smurthys:
Seems like we should not have included `index.md`: the version just added interchanges "site" and "wiki". We need to keep the structure the same as that in the docs for 2.0 presently deployed.

srrollo:
Thanks for catching that, it should be fixed now. IIRC, `index.md` and `Home.md` in the repo are both distinct from their wiki counterparts.

PR #256, opened 2018-08-12
Move server-version comparers to own script

### LINKED ISSUE #255: Server-version comparison functions are too long to be part of another script (E)
Functions that compare server versions are presently in `addHelpersCore.sql` though the comparison functions [span enough number of lines](https://github.com/DASSL/ClassDB/blob/caf58e759f23e56363130b95a021f56937655e5f/src/db/core/addHelpersCore.sql#L601-L758) to be in their own script.

This change requires a corresponding change to test scripts.

smurthys:
The commits in this PR make the following changes:
- Move functions that compare server-versions to their own script
- Move unit tests for server-version comparers to their own script
- Add the new script to db-core installer

This change is being done in preparation for server-version specific code as part of Milestone 4 (M4).

Fixes #255


smurthys:
Thanks for the reviews.

PR #261, opened 2018-08-15
Reject row deletion over tables DDLActivity and ConnectionActivity

### LINKED ISSUE #176: Rows can be selectively deleted in DDL and Connection activity tables (M)
Rows in tables DDLActivity and ConnectionActivity can be selected deleted, though that can only be done by the ClassDB role or by a super user.

smurthys:
The commits in this PR make the following changes:
- Add triggers to prevent row deletion over tables `DDLActivity` and `ConnectionActivity`. The ClassDB role or a superuser can still truncate these tables.
- Update test script for the new triggers. Also fixes a bug in the part that tested existence of triggers: used to test if at least one of the triggers are defined, instead of testing if all expected triggers are defined.

In future we should also consider preventing direct truncation of these table. For that we need to also implement functions to initiate truncation so the tables can be somehow purged. However, this functionality needs to be discussed as part of a larger log-management feature set.

Fixes #176

### DIFF
--- a/src/db/core/addUserMgmtCore.sql
+++ b/src/db/core/addUserMgmtCore.sql
@@ -247,7 +247,7 @@ $$;
 
 
 --Define a trigger function to raise an exception that an operation is disallowed
--- used to prevent non-user inserts to, and any updates to, tables DDLActivity
+-- used to prevent non-user insert and any update or delete over tables DDLActivity
 -- and ConnectionActivity
 -- trigger functions cannot accept parameters, but the 0-based array TG_ARGV will
 -- contain the arguments specified in the corresponding trigger definition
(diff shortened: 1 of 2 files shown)


afig:
The changes look great, thanks @smurthys. Tests pass and `DELETE`s are correctly blocked on `ClassDB.DDLActivity` on 9.4 and 10. I only see one minor issue with the comment on L36 of `testUserMgmt.sql`.

I agree that direct truncation should eventually be blocked as well, and that implementing that should require ClassDB to have additional log management features that allow for similar functionality.

smurthys:
Thanks @afig for the review. I just corrected the comment.

smurthys:
I just updated the [docs page](https://github.com/DASSL/ClassDB/wiki/Activity-Logging) to say the rows in the log tables cannot be directly manipulated. While there, I also edited that page for correctness, completeness, etc.

smurthys:
Thanks for the reviews

Issue #229, opened 2018-06-17
Guard uses of GRANT: changed in pg9.5 (N)


KevinKelly25:
Using postgreSQL version 9.3, `initializeDBCore.sql` has a syntax error on line 95. As shown here:
```psql
psql:core/initializeDBCore.sql:95: ERROR:  syntax error at or near "CURRENT_USER"
LINE 1: GRANT ClassDB TO CURRENT_USER;
```

It looks like in version 9.5 the syntax for using GRANT changed as seen [here](https://www.postgresql.org/docs/9.3/static/sql-grant.html) by comparing versions.

afig:
The change in the syntax only affects direct use of `CURRENT_USER` or `SESSION_USER`. These `role_specifications` can still be used if the statement is dynamically generated. Our biggest use of `CURRENT_USER` is as a default parameter. This use is unaffected by the change in syntax.

The change in syntax also affects various other commands, including `REVOKE`, `ALTER ROLE`, `DROP OWNED BY`, and `REASSIGN OWNED BY`

There is only one statement in ClassDB's source that is affected (the same one pointed out above). There are 6 more affected uses of `CURRENT_USER` in the test scripts (5 in `testDisallowSchemaDrop.sql` and 1 in `testRoleBaseMgmt.sql`.

A potential workaround is converting the affected uses of `CURRENT_USER` into dynamic statements, which will result in Postgres calling `CURRENT_USER` before generating the statement, avoiding syntax issues. Since they involve the use of the `EXECUTE` statement and `FORMAT()` function, they can only be called in a `plpgsql` block. For example, the issue above can be mediated by replacing the affected code on L95 with the following:

```sql
--Grant ClassDB to the current user
-- allows altering privileges of objects, even after being owned by ClassDB
DO
$$
BEGIN
   EXECUTE FORMAT('GRANT ClassDB TO %s', CURRENT_USER);
END
$$;
```

Of course, this comes with all the drawbacks of executing a dynamically generated statement (no caching, increased difficulty in debugging, and no syntax highlighting in a text editor).

Once that one use of `CURRENT_USER` has been updated, ClassDB can be installed without issues caused by the change in syntax. Modifying the 5 instances in the test scripts allows them to run (they do not pass due to an unrelated issue). I do not foresee any other issues cause by this change.

afig:
[...]

PR #265, opened 2018-08-15
Guard uses of ALTER SYSTEM for pg9.3 compatability

### LINKED ISSUE #228: Guard uses of ALTER SYSTEM: changed in pg9.4 (N)
This issue is based on a [finding](https://github.com/DASSL/ClassDB/issues/225#issuecomment-397840320) by @KevinKelly25.

`ALTER SYSTEM` was introduced in Postgres 9.4 and thus its execution should be guarded by the server version number. It appears that prior to Postgres 9.4, it is necessary to change postgresql.conf to get the same effect as using `ALTER SYSTEM`.

`ALTER SYSTEM` is presently used in `enableConnectionLoggingReco.psql` and `disbleConnectionLoggingReco.psql`.

afig:
This PR adds a guard that prevents any statements located in `enableconnectionLoggingReco.psql` and `disableConnectionLoggingReco.psql` from being run if the Postgres server version is < 9.4. This guard is needed because the `ALTER SYSTEM` statements were not available prior to pg9.4, and these settings need to instead be manually set in the `postgresql.conf` file. Also note that because of how the guard is implemented, it is no longer possible to use the `addAllToServer.psql` script to add everything to the server [when using Postgres 9.3]* (which is okay since there is only one other server file apart from these two).

\*Edited to mention that the issue with `addAllToServer.psql` only applies to 9.3

As mentioned in the script comments, ClassDB's server comparers are not accessible. An equivalent call to one of these functions is included in a comment to make locating the backwards-comparable alternative easier.

Currently, I believe it makes more sense to include details about manually setting these server parameters in [ClassDB's Setup page](https://github.com/DASSL/ClassDB/wiki/Setup). However, please feel free to suggest an alternative location.

Closes: #228

### DIFF
--- a/src/server/reco/assertAlterSystemAvailability.sql
+++ b/src/server/reco/assertAlterSystemAvailability.sql
@@ -36,8 +36,7 @@ BEGIN
          MESSAGE = 'could not set logging-related settings',
          DETAIL = 'Logging-related settings must be set manually in Postgres'
                   ' 9.3 and earlier.',
-         HINT = 'See https://dassl.github.io/ClassDB/Setup for more'
-                ' information.';
+         HINT = 'Consult ClassDB documentation for more information.';
    END IF;
 END;
 $$;
--- a/src/server/reco/assertAlterSystemAvailability.sql
+++ b/src/server/reco/assertAlterSystemAvailability.sql
@@ -21,8 +21,6 @@
 -- postgresql.conf in pg9.3 or below. This script and any invocations of this
 -- script should be removed when support for pg9.3 is dropped.
 
-\set ON_ERROR_STOP on
-
 DO
 $$
 BEGIN
--- a/src/server/reco/assertAlterSystemAvailability.sql
+++ b/src/server/reco/assertAlterSystemAvailability.sql
@@ -1,0 +1,43 @@
+--assertAlterSystemAvailability.sql - ClassDB
+
+--Andrew Figueroa
+--Data Science & Systems Lab (DASSL)
+--https://dassl.github.io/
+
+--(C) 2018- DASSL. ALL RIGHTS RESERVED.
+--Licensed to others under CC 4.0 BY-SA-NC
+--https://creativecommons.org/licenses/by-nc-sa/4.0/
+
+--PROVIDED AS IS. NO WARRANTIES EXPRESSED OR IMPLIED. USE AT YOUR OWN RISK.
+
+--This script does not need to be run individually for an installation of
+-- ClassDB and does not modify the server or any database. It is called by
+-- other installation scripts when it is needed
+
+--This script asserts that the ALTER SYSTEM command is available for use by
+-- verifying that the current server version is not <9.4
+
+--Any configuration options that are set by ALTER SYSTEM must be set manually in
+-- postgresql.conf in pg9.3 or below. This script and any invocations of this
+-- script should be removed when support for pg9.3 is dropped.
+
+\set ON_ERROR_STOP on
+
+DO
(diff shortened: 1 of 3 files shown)


smurthys:
Thanks @afig for the contribution. These changes are inline with the team's decision on how to resolve this issue.

I have two suggestions:
- Factor out the new code that asserts the current server version to a separate script file such as `assertAlterSystemAvailability.sql` and invoke it from the existing scripts using the `\ir` meta-command. This approach reuses code and improves maintainability.
- Set the exception's hint to something more generic such as `Consult ClassDB documentation for details` so we don't need to change the code if/when the docs are reorganized.

afig:
Thanks for the suggestions @smurthys. I agree with both parts and have pushed corresponding changes.

smurthys:
Nice changes. I recommend moving the meta-command out of `assertAlterSystemAvailability.sql' into each of the calling script so that:
- the calling script can decide what happens on error
- the assertion script is SQL
- the assertion script is easily made more reusable later: I can imagine reusing this script to assert any pg version and customize all parameters of the exception (which will then make the assertion script psql, but that will be OK)

smurthys:
@afig Would you please add a brief note to the Activity Logging docs page about altering connection settings in pre9.4 pg?

afig:
Thanks again for the suggestion. The use of a meta-command in a `.sql` script should have sounded some alarms before I pushed the changes.

I had added a note to the Setup page about altering the connection settings. However, I think it makes more sense to have it under the Activity Logging page as suggested and add a link to it on the Setup page.

smurthys:
Thanks @afig for the changes. I like the note on the setup page and the detailed note on the logging page. (FYI, I tinkered with those pages for consistency and clarity.)

I like clarification on two things: Have you run `addAllToServer.psql` on pg9.3 and verified the script stops after showing the exception.

Related, would you please expand on the following text in the PR description? Is it saying `addAllToServer.psql` can no longer be used with any pg version? I hope not. 

> Also note that because of how the guard is implemented, it is no longer possible to use the `addAllToServer.psql` script to add everything to the server (which is okay since there is only one other server file apart from these two).

afig:
On my installation with Postgres server version 9.3.24, both `psql` client versions 9.3.24 and 10.5 correctly show the exception and stop when `addAllToServer.psql` is run. Running either client version with Postgres server 9.4.19 results in the script completing successfully (with an unrelated patch to workaround issue #229).

The PR description should have been clearer. `addAllToServer.psql` can no longer be used with Postgres version 9.3 or below, but it can still be used with versions 9.4 or higher.

smurthys:
Thanks @afig for the clarification. I apologize but I am now a bit confused by the first paragraph in your recent comment:

> On my installation with Postgres server version 9.3.24, both psql client versions 9.3.24 and 10.5 correctly show the exception and stop when addAllToServer.psql is run. 

Do you really mean the changes cause an exception and stop in pg10.5? They should not.

afig:
That was referring to `psql` _client_ versions 9.3.24 and 10.5 both connecting to a Postgres 9.3.24 server.

smurthys:
Ah, psql version 10.5. Thanks I notice it now. Sorry I insisted on your reply.

smurthys (commented):
These changes are ready to merge

smurthys (approved):
These changes are ready to merge.

KevinKelly25 (approved):
Thank you @afig for this guard. This works well on version 9.4 and 10.

afig:
Thank you for reviews everyone.

PR #266, opened 2018-08-15
Guard use of pg_signal_backend

### LINKED ISSUE #225: Guard uses of role pg_signal_backend: changed in pg9.6 (N)
Making ClassDB compatible with Postgres versions prior to 9.6 (back up to 9.3 to be specific) requires performing the following operations only if the server role `pg_signal_backend` exists:
- In `initalizeServerCore.sql`, grant `pg_signal_backend` to role `ClassDB`
- In `addConnectionMgmtReco.sql`, make `ClassDB` the owner of function `kilUserConnections`

But for the comments in code about the conditional execution, no other documentation change should be necessary.

smurthys:
The changes in this PR guard the use of the predefined server role `pg_signal_backend` only in versions 9.6 or later. Fixes #225 

**Note:** Using the server version to address this issue is technically incorrect because the functionality in question is determined by the existence of a specific server role which can be tested in the `pg_roles` table. However, testing for server version is more practical and maintainable, and we should be able to trust no one has removed a server role. (We make many such assumptions about the environment.)

### DIFF
--- a/src/db/reco/addConnectionMgmtReco.sql
+++ b/src/db/reco/addConnectionMgmtReco.sql
@@ -77,15 +77,24 @@ $$ LANGUAGE sql
    SECURITY DEFINER;
 
 --Change function ownership and set execution permissions
-ALTER FUNCTION
-   classdb.killConnection(INT)
-   OWNER TO ClassDB;
-REVOKE ALL ON FUNCTION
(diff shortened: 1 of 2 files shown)


smurthys:
Thanks for the reviews.

Issue #267, opened 2018-08-16
Server-version comparers are not marked immutable (M)


smurthys:
The comparison functions in `addServerVersionComparersCore.sql` are immutable but are not marked as such.

afig:
Fixed in #268

PR #268, opened 2018-08-16
Use machine-readable version number to determine current server's version

### LINKED ISSUE #263: Server version comparison uses incorrect preset option (W)
The [Postgres documentation related to getting version information](https://www.postgresql.org/docs/10/static/functions-info.html#id-1.5.8.30.46), contains the following statement (emphasis mine):
 
>`version` returns a string describing the PostgreSQL server's version. You can also get this information from `server_version` or for a machine-readable version, `server_version_num`. **Software developers should use `server_version_num` (available since 8.2) or PQserverVersion instead of parsing the text version**.
 
The machine-readable `server_version_num` returns an integer containing a numerically sequential version number. As examples, for Postgres version 9.3.24, this number is `90324`, for Postgres 9.6.10, it is `90610` and for version 10.5, `100005`.
 
Currently in `addServerVersionComparersCore.sql`, we are using `server_version`, rather than the recommended `server_version_num`. As a matter of fact, the variability of the non-machine-readable version caused quite a few issues when getting [PR 232](https://github.com/DASSL/ClassDB/pull/232) ready to merge. In the end, we ended up with more functionality than if we had gone with the more direct route (can optionally ignore minor versions and input version numbers as the standard x.y.z rather than xxyyzz).
 
ClassDB should internally implement use of `server_version_num` to avoid the variability of the `server_version` string. Currently, if a distribution adds branding before the version number in the version string, our code will no longer function correctly (we already handle the possibility of there being branding after the version number). Our current internal functions should still take in x.x.x/x.x version numbers for convenience and to avoid having to make widespread API-breaking changes.

smurthys:
This PR makes the following changes:
- Add a new function `ClassDB.intServerVersionToString` to convert a machine-readable pg server version number to string
- Use the server setting `server_version_num` instead of 'server_version' in function `ClassDB.getServerVersion`
- Remove support for distro suffix in function `ClassDB.compareServerVersion`
- Mark all functions in the script `addServerVersionComparersCore.sql` as `IMMUTABLE`
- Add test for new functionality
- Remove test for version numbers with distro-suffix

Fixes #263 #267

### DIFF
--- a/src/db/core/addServerVersionComparersCore.sql
+++ b/src/db/core/addServerVersionComparersCore.sql
@@ -33,17 +33,48 @@ END
 $$;
 
 
+--Define a function to convert an integer pg server version number to string
+--https://www.postgresql.org/docs/10/static/libpq-status.html#LIBPQ-PQSERVERVERSION
+CREATE OR REPLACE FUNCTION ClassDB.intServerVersionToString(v INT)
+RETURNS VARCHAR AS
+$$
+BEGIN
+   IF v >= 100000 THEN
+      --pg10 or later has 2-part numbers such as 10.1 and 11.0, which are encoded
+      -- as numbers with 6 or more digits such as 100001 and 110000
+      RETURN CONCAT((v/10000)::VARCHAR, '.', (v%10000)::VARCHAR);
+   ELSE
+      --pre-pg10 versions are 3-part numbers which are encoded as 5-digit numbers
+      -- with two digits used (0 filled) for each part
+      -- e.g., 9.1.5 and 9.2.0 are encoded as 90105 and 90200 respectively
+      RETURN CONCAT((v/10000)::VARCHAR, '.', (v%10000/100)::VARCHAR, '.',
+                    (v%100)::VARCHAR
+                   );
+   END IF;
+END;
+$$ LANGUAGE plpgsql
+   IMMUTABLE
+   RETURNS NULL ON NULL INPUT;
+
+ALTER FUNCTION ClassDB.intServerVersionToString(INT) OWNER TO ClassDB;
+
+
 
 --Define a function to get the server's version number
---Removes additional info a distro may have suffixed to the version number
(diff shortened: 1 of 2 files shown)


smurthys:
I chose to convert current server's version from int to string for two reasons (instead of converting the input string version number to int):
- Only one function needs to change
- Considerably more logic is needed to ignore parts during comparison if int versions are to be compared. (need to bit-mask parts not of interest). This is compounded by the change in numbering starting pg10.

KevinKelly25:
Thank you @smurthys for adding this useful function. Everything works well on 9.4 and 10. 

The only small thing that I noticed was the use of `v` as a variable. While it is rather easy to guess the meaning of v, there aren't many places, if any, in ClassDB where it isn't immediately evident what a variable means. With `v` it took me a few seconds to make sure I knew the meaning. To me it doesn't fit into the nomenclature of ClassDB where it is normally easily readable. It seems like using `version` instead of `v` would fit better. However, it might be just me overthinking this and it is worth a small discussion.

afig:
Looks great, thanks for updating the implementation @smurthys. The functions work as expected and tests pass on versions 9.6 and 10, and converting from int to string is certainly ideal all things considered.

I have two minor suggestions:

- Add a test in `testServerVersionComparers.sql` that uses the old versioning scheme but with a `0` second part (for example, asserting that `ClassDB.intServerVersionToString(90005) = '9.0.5'`). It seems like such a test will pass without issue, but the case is distinct enough that it might warrant a test.

- The comment on L68 of `addServerVersionComparersCore.sql` seems unfinished (ends with 'and` and does not point to a second source of information on the next line).

I also agree with @KevinKelly25's suggestion about replacing `v` with a more descriptive variable name.

smurthys:
Thank you for the review. It seems like I forgot to replace the variable name I had used in the version I had developed in pgAdmin. Thanks for catching that.

I just pushed changes that address the three issues pointed out. Because a parameter name changed, please run the following query before running the revised script. This step is **not** required in production:

```sql
DROP FUNCTION ClassDB.intServerVersionToString(INT);
```

PR #271, opened 2018-08-17
Add Guard for DROP OWNED BY

### LINKED ISSUE #269: Guard uses of DROP OWNED: changed in pg9.5 (N)
The names CURRENT_USER and SESSION_USER in DROP OWNED is permitted only from pg9.5. Thus, this pattern of code needs to be guarded and executed as dynamic SQL in pg9.4 and earlier.

The script `testDisallowSchemaDrop.sql` has a few instances that need guarding.

KevinKelly25:
This PR adds a temporary function that when given a `DROP OWNED BY` query it will dynamically execute the query with an attached `CURRENT_USER`. All uses of `DROP OWNED BY CURRENT_USER` have been replaced with this temporary function with the replaced query as the parameter. This also maintains the purposefully odd case of `DrOP oWNeD BY ` on Ln120.

Fixes #269

### DIFF
--- a/tests/testDisallowSchemaDrop.sql
+++ b/tests/testDisallowSchemaDrop.sql
@@ -33,7 +33,7 @@ $$;
 -- 9.4 and lower versions need dynamic queries to work with CURRENT_USER in
 -- DROP OWNED BY queries. Remove this function once support for pg9.4 is dropped
 -- and use the following 'DROP OWNED BY CURRENT_USER' and its variations.
-CREATE OR REPLACE FUNCTION pg_temp.doDropOwnedByCurrentUser(query VARCHAR) RETURNS VOID AS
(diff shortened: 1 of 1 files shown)


afig:
Changes look good and follow the recommend implementation. I only noticed some minor comment issues:

- L32: quieries -> queries

- L43: Add a separator (`----------------...`) here might not be necessary since it is not inside of the function and thus is already visually separated through indentation.

smurthys:
Thanks @KevinKelly25. Just one minor improvement: change parameter name `query` to `queryHead` because its value is only the beginning of a query, rather than the entire query.

(I have yet to test the changes.)

KevinKelly25:
@afig Thank you for the the suggestions. I think you and @smurthys posted around the same time and I missed your comment somehow. I apologize for that. I fixed this issues you mentioned.

KevinKelly25:
Thank you for the reviews

PR #274, opened 2018-08-17
Fix removeAllFromDB.sql

### LINKED ISSUE #225: Guard uses of role pg_signal_backend: changed in pg9.6 (N)

KevinKelly25:
Added a drop function to `removeAllFromDB.sql` which incorporates a drop for `ClassDB.killConnection(INTEGER)` which was previously breaking the script. 

Fixes #225

### DIFF
--- a/src/db/removeAllFromDB.sql
+++ b/src/db/removeAllFromDB.sql
@@ -103,6 +103,8 @@ DROP FUNCTION IF EXISTS ClassDB.allowSchemaDrop();
 
 DROP FUNCTION IF EXISTS ClassDB.getSessionID();
 
+DROP FUNCTION IF EXISTS ClassDB.killConnection(INTEGER);
+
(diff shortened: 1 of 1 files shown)


KevinKelly25:
Thank you for pointing that out @smurthys . I changed it to better fit the nomenclature.

KevinKelly25:
Thank you for reviews

PR #275, opened 2018-08-18
Added M4 docs to repo

srrollo:
Adds he M4 docs to the repo


srrollo:
Thanks @afig, I just added that change to the PR

PR #281, opened 2019-08-12
Clarify comment on granting login privilege

### LINKED ISSUE #280: Roles with login are not granted login privilege
If function `createRole` is called for a role that already exists, the logic to grant login privilege to the role incorrectly skips granting that privilege.

The issue is due to an [incorrect logical expression](https://github.com/DASSL/ClassDB/blob/e0db7e0e9519673d5f4320f59fa7d5d60bec5e73/src/db/core/addRoleBaseMgmtCore.sql#L315-L317) which reads as follows:

```sql
   IF NOT($3 OR ClassDB.canLogin($1)) THEN
      EXECUTE FORMAT('ALTER ROLE %s LOGIN', $1);
   END IF;
```
The code should be as follows:

```sql
   IF NOT $3 OR ClassDB.canLogin($1) THEN
      EXECUTE FORMAT('ALTER ROLE %s LOGIN', $1);
   END IF;
```

smurthys:
This commit revises the comment associated with the code to grant login privilege when a role is created. Fixes #280

Only comment change; **no code change**.

### DIFF
--- a/src/db/core/addRoleBaseMgmtCore.sql
+++ b/src/db/core/addRoleBaseMgmtCore.sql
@@ -311,7 +311,7 @@ BEGIN
 
    --in case a pre-existing server role is now registered, give that role LOGIN
    -- capability if it is a user (in case that privilege was somehow removed);
-   -- but don'o't remove LOGIN from a team: instr. may have reason to let a team login
+   -- but don't remove LOGIN from a team: instr. may have reason to let a team login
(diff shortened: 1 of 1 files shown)


smurthys:
I just pushed a change that fixes a spelling error in the comment explaining the logic. I also opportunistically changed a phrase in another comment. Please review

KevinKelly25 (approved):
Thank you for updating the comments to make them more understandable. I don't see any errors

Issue #277, opened 2018-09-02
Dropping a user as student from a DB removes the user from student role in all DBs on the server


KevinKelly25:
If you drop a student from one ClassDB database it drops that student from all ClassDB databases. It makes sense why it does that looking at `dropStudent` function but it seems inconsistent with `createStudent`. For example:
 
If I have two databases cs305 and cs205. If I add a student to cs205 that student will not show up in cs305. However, if I add a student to both cs205 and cs305 then drop the student from either of the classes it will remove the student from both classes. At the very least not letting them connect to either and also removing them from student activity table of both.

afig:
If I understand the situation correctly, in the example/scenario provided, the student should still be in the `ClassDB.RoleBase` table of the other class (but will not show up in the `ClassDB.Student` view, since that view checks for the `ClassDB_Student` group role, which the student no longer has).

The main issue is that the `ClassDB_Student` group role is revoked even if that user is still a student in another database within the same Postgres server cluster.

I believe related behavior has been informally discussed in-person before, but not in the context of being an issue.

KevinKelly25:
I believe you are right about what is happening, especially after doing a lot of tests surrounding the issue.
 
Maybe we could add a function to remove a student from just a single database? I think it would be relatively simple, the only problem I see with that is testing that functionality in the test scripts, since you would have to be testing in multiple databases.

smurthys:
Ideally function `dropXYZ` we should revoke a dropped user's membership from a ClassDB role only if no the user does not have that role in any other DB. However, that is not (easily) possible in Postgres due to limitations on cross-DB queries.

An easy solution is to add an optional parameter such as `dropServerRole` to `dropXYZ` functions which defaults to `FALSE`. We can drop the server role only if the new parameter is `TRUE`. This approach has the disadvantage of leaving zombie server roles, but I don't see that being too much of an issue. It is possible a future `createXYZ` fails due the role remaining, but that too does not seem much of an issue. I can expand on these concerns later as we develop the solution.

Testing with multiple DBs should not be an issue because a psql script can change connection using the `\connect` meta-command.

EDIT: Finishing the apparently unfinished sentence from many moon ago. Apologies for the delay in noticing it.

[...]

PR #282, opened 2019-08-12
Limit database connection only to users known to that database

### LINKED ISSUE #278: Users known in one ClassDB database are able to log in to all ClassDB databases on the same server (W)
With two ClassDB databases on the same server, and `user1` is added as a ClassDB role (say as a student) to one database, but the same user is not added as a ClassDB role to the other database, `user1` is still able to login to the second database. 

In this scenario, a student user does not have unauthorized access to any object in the second database, but instructors and DB managers could.

I will add another comment soon with an analysis of the problem and potential solutions.

This issue is related to Issue #277.

smurthys:
The commits in this PR remove automatic DB connection to ClassDB group roles (at DB initialization) and instead manage that permission separately for each login role: grants permission in function `createRole`; revokes permission in function `revokeClassDBRole` if the role has no more ClassDB roles.

These commits fix #278.

This fix can be removed when Issue #277 is fixed, assuming that issue is addressed by creating database-specific names for ClassDB group roles as [proposed in a comment](https://github.com/DASSL/ClassDB/issues/277#issuecomment-417973035) at that issue.

The changes are tested manually. Privilege tests need to be updated.

### DIFF
--- a/src/db/core/addRoleBaseMgmtCore.sql
+++ b/src/db/core/addRoleBaseMgmtCore.sql
@@ -316,6 +316,16 @@ BEGIN
       EXECUTE FORMAT('ALTER ROLE %s LOGIN', $1);
    END IF;
 
+   --permit the role to connect to this database
+   -- can remove this code segment if db-specific group roles are used (when
+   -- initializing the database) to address Issue #277
+   --this code segment is intentionally not merge/optimize with the preceding
+   -- segment (which grants LOGIN to a user role) for clarity, and to make it
+   -- easier to remove/modify this code when Issue #277 is addressed
+   IF ClassDB.canLogin($1) THEN
+      EXECUTE FORMAT('GRANT CONNECT ON DATABASE %I TO %s', current_database(), $1);
+   END IF;
+
 
    -------- schema management --------------------------------------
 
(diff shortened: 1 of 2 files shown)


KevinKelly25:
All unit and privilege tests pass for me. I also tested this change with role in multiple ClassDB instances and each role was only able to connect to the databases it was added to (as expected). The only small issue which we already discussed in our meeting is that there is a misspelling in a comment on line 314 of `addRoleBaseMgmtCore.sql`. 

Thank you for implementing this temporary solution @smurthys

smurthys:
Thanks for the review and for testing @KevinKelly25.

I see one issue in `initializeDBCore.sql`: In addition to [**not** granting database connection](https://github.com/DASSL/ClassDB/pull/282/files#diff-f03b0a5cd148bcf00738fd90129bea9bR91-R93) to ClassDB group roles, it is also necessary to revoke database connection from those roles. The revocation is necessary if an existing ClassDB installation is updated with this revision. The downside to the revocation is that existing users immediately lose the ability to connect, but that can be addressed by explicitly granting db connection to existing users. I am thinking about a simple way to do that, and I welcome suggestions.

KevinKelly25:
One possible approach is that we can run the following if the `RoleBase` table exists already: 
```sql    
    FOR User IN
      SELECT RoleName FROM ClassDB.RoleBase
    LOOP
       EXECUTE FORMAT('GRANT CONNECT ON DATABASE %I TO %s', current_database(), User);
    END LOOP;
```
The problem with this approach is that this will also be run if user already can connect to the database. I do not believe this causes an error (worth checking though) so it is more of an efficiency problem.

smurthys:
Thanks for the tip @KevinKelly25. I played around with a similar solution. The query needs to be over pg_roles table if executed in `initializeDBCore.sql`. which is where I think the solution should be. I will think more about over the next couple of days.

KevinKelly25:
I agree it logically should be in `initializeDBCore.sql`. However, I do not believe it is possible to use `pg_roles` for this operation as it shows roles for the whole server, not just the database it is currently in. Since role information is stored at the server level rather then the database you would have to relate the role to a specific database. I am not sure that is possible with the system tables since they don't have database specific roles in the current setup.

smurthys:
Indeed. `RoleBase` and `pg_roles` would need to be joined.

